PR #24, opened 2015-02-20
Add XML support

immewnity:
This adds XML support. By default, most browsers will show XML due to the Accept header sent. curl will show JSON by default. Either of these can be changed by appending ?format=json or ?format=xml onto the URL, respectively. Changing this behavior would require changing browser or curl settings.


phalt:
I'm still apprehensive to make this change as we'll be breaking a lot of web-based services using the API (such as the javascript widget on the front page of pokeapi.)

As a rule: once a web API is in the public domain you should not drastically change it's functionality. We need to investigate this more and see if we can make it json default and still support xml.

Also, this will need tests to make sure 1) xml works and 2) json is still the default.

immewnity:
Maybe with v2?

Sent from my Windows Phone
________________________________

phalt:
Definitely for V2 - the V2 api will be using Django REST framework, which does default to JSON and still supports XML

immewnity:
Theoretically... the XML change could be, say v1.5, making the pokeapi.co URL http://www.pokeapi.co/api/v1.5/. In .htaccess, it wouldn't be too hard to redirect any requests to /api/v1/* to /api/v1.5/*?format=json, with * being anything after /api/v1/.

phalt:
I don't want to get into *.5 versions of the API. In fact, I regret making the version explicit in the URL - I'd have rather just accepted an http header but I was naive about versioning when I originally created PokéAPI.

Are you certain you can't find any JSON to XML parser? I can't believe this doesn't exist for the tools you're using.

immewnity:
None that work all that well, from what I've read. I'll keep looking, though.

This would be more of a temporary fix until v2, though, no? v2 could then be just /api/.

Date: Fri, 20 Feb 2015 08:01:41 -0800

Issue #9, opened 2014-12-04
CSV to database builder


phalt:
We need to implement a new way of taking CSV data and turning it into database tables, whilst still ensuring the V2 models can communicate with the data correctly.

vamsivarikuti:
Can you be more specific about this task. I like to work on it.

zaneadix:
I second that. I see where you mention your previous implementation wasn't ideal. How should things change? Would it be efficient enough to just consolidate everything into a single script rather than individual functions?

phalt:
There is a rich collection of data over on Veekun's [pokedex](https://github.com/veekun/pokedex), which is all exported to csv.

Veekun's pokedex transforms this all into a mysql (I think?) database at run time. Ideally we'd want a similar script that powers through all those scripts and maps them to the models that we have in PokéAPI. They are slightly different, with slightly different relationships. This is again due to my initial design, which was nothing more than 5 minutes of "that'll do" to get the relationships to work (remember PokeAPI was a weekend project that span out of control :D ).

It might be a good idea to build some abstract models that map to Veekun's CSV files better but still allows the old V1 API to work nicely.

aindurti:
I agree with CSV approach. Querying it is much faster, and the format is way more lightweight for large sets of data. It removes repetitive keyframes, and although development on it can be a little challenging because there isn't much standardization, the API itself will take care of that.

Mapping with the old V1 API might be easier as well, provided we upgrade the returned object to something more business-logicky (more relational/tabular). Try using ODATA, a version of JSON that allows for more powerful querying for large data-sets.


The hard part of mapping with the old API will be allowing conditional inputs to the requests that allow for better filtration. See issue #28

phalt:
@aindurti Just to be clear - I don't want to make CSV files the datastore here, I want to use them to generate the datastore on the server by creating database rows. Querying CSV files will put a lock on that file while it is being read. This is undesirable considering the amount of traffic PokéAPI receives, it will cause broken requests.

zaneadix:
Pull request #35 should hopefully take care of this.

matthewdias:
This might be out of context because I haven't been following this project very closely, but as I understand, V2 is supposed to map closer to the models used by Veekun. In that case, many database apps provide export from SQL to JSON. Could that be of any use here if we were to pull the database straight from Veekun?

zaneadix:
Taken care of in pull request #40

Brisingr211:
[...]

PR #25, opened 2015-02-22
Initial Django REST framework progress

phalt:
Currently a WIP


vamsivarikuti:
Hi phalt, can you merge this pull request into master. I'm free today. I'll checkout the "csv to database" task #9. The database should be accessed as an api through django-rest-framework right?

phalt:
@vamsivarikuti this is no where near being completed :P

phalt:
Updated to reflect new model changes. Not yet complete.

zaneadix:
Hey @phalt has this actually been working for you so far? I've been tinkering with this and feel like I have things wired up right for Abilities and no matter what I do when I visit http://localhost:8000/api/v2/ability/ I get a 500 page.

phalt:
@zaneadix sorry no. As I keep saying - I'm building APIs full time at work and development on this has all but stopped :cry: I committed this half done, so it doesn't surprise me there are 500's. Turn on DEBUG=True and that'll point in the right direction!

zaneadix:
Thanks for getting back. I found out it was actually just a few missing commas in the REST_FRAMEWORK settings. I've got Abilities serving now. Working on more.

Issue #8, opened 2014-12-04
Build V2 models to reflect csv in data/v2/csv


phalt:
The V2 resources should map directly to the csv files in ```data/v2/csv```, to ensure easier mapping data and especially when we need to update the models.

zaneadix:
Have you made a lot of progress on this task? I'm not a Django guy t all really so I've been playing around with this to get down the Model layer and I'm wondering if I could be any help to flesh it out legitimately.

phalt:
@zaneadix I got an awful lot of emails from you last night, testing things? Please remove my email next time.

I've made almost no progress on this as I'm quite busy. I made this open source so people could contribute but alas, it seems no one actually wants to now they can :D You're free to make a start.

zaneadix:
Sorry again about those dumb emails. Just wanted To give an update so anyone who wants to jump in can message if they want to. I'm also putting some time into the database build script just to test the models out. I've got the abilities in place and am going to go through to find where I can abstract similar models to streamline things.

phalt:
@zaneadix wow nice, thanks! This would be amazing if you could. Feel free to open up a Pull Request and I can review the work you've done so far and offer advice.

zaneadix:
Pull request #35 should hopefully knock this issue out. Some more data can be pulled in from veekun (Items, Berries, Regions) but up to this point there should be models for everything else.

zaneadix:
Taken care of in pull request #40

PR #33, opened 2015-04-07
Base Model Implementations

zaneadix:
So up to this point I've created models and filled tables for all but Pokemon and Moves. I didn't want to dig in really far just to find out I've been doing stuff dumb. Im not a Django, Python or even SQL guy really (mostly node/mongo for me). I'm digging into learning this stuff purely cause I want this thing to live.

Theres mainly just changes to pokemon_v2/models.py and data/v2/build.py.

I may have gone a bit over board with the base model abstractions but I figured it was better than rewriting the same field declarations over and over.

To check the models in the table I've been going into the Django shell and running

> execfile('data/v2/build.py')

which is clearing and rebuilding each table every time it runs. It's creating an issue where the 'id' field index is not being reset so tables are not indexed from 1 currently (unless I explicitly enter the 'id' value on Model instantiation which I know isn't a great thing to do either).

I'm wondering if you might have already had some idea of how to do the build script in a more abstract and concise way.

Also, I have been using ForeignKeys to map between each of the tables id fields. As far as I know I could really just be creating IntegerFields but the ForeignKeys seemed to be more descriptive when reading the code. I'm not sure if this is an unconventional or wasteful thing to do so if you have any hard opinions on this I'll take them to heart and clean up the code wherever I can. I see that you've been using them to create small "mapping tables" between Pokemon and their abilities, etc. Is this the better thing to do?

One last thing, I was thinking about pulling in some of the other data from veekun (like berries) because natures and other things map to berry flavors and feel like we should avoid loose ends if possible. You think this is worth doing?

Hope all looks good to you so far. All advice is appreciated.


config/settings.py
```diff
@@ -83,7 +83,7 @@
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
         'NAME': 'pokeapi_co_db',
-        'USER': 'root',
+        'USER': 'zane',
```
phalt: Please remove this, it has to be set to root for the live database.

config/settings.py
```diff
@@ -8,7 +8,7 @@
 TEMPLATE_DEBUG = DEBUG
 
 ADMINS = (
-    ('Paul Hallett', 'paulandrewhallett@gmail.com'),
+    ('Zane Adickes', 'zaneadix@gmail.com'),
```
phalt: I'd suggest removing this too, your email will get a /lot/ of log errors otherwise.

pokemon_v2/migrations/0001_initial.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
phalt: There are a /lot/ of migrations here, you can squash them using the `squashmigrations` command found here: https://docs.djangoproject.com/en/1.7/topics/migrations/#squashing-migrations

phalt:
This looks amazing! Comments coming but I wanted to say thank you first :gift: :tada:

phalt:
> Im not a Django, Python or even SQL guy really 

Honestly you've done a pretty great job so far

> may have gone a bit over board with the base model abstractions

This provides so much more flexibility, I'd keep it that way.

> I have been using ForeignKeys to map between each of the tables id fields. As far as I know I could really just be creating IntegerFields but the ForeignKeys seemed to be more descriptive when reading the code. I'm not sure if this is an unconventional or wasteful thing to do so if you have any hard opinions on this I'll take them to heart and clean up the code wherever I can.

ForeignKeys are presented in the database as foreign keys too, so we should definitely use them as they provide relations for us. The ORM intelligently uses them.

> One last thing, I was thinking about pulling in some of the other data from veekun (like berries) because natures and other things map to berry flavors and feel like we should avoid loose ends if possible. You think this is worth doing?

If you think the scripts can be built for this, then go for it.

### Final notes

This is great and I will merge it into master but the API portion still needs to go live. I will work on that, now that the models are there, the API should be easier to build.

I do want the following things done:

- [ ] Squashmigrations to fewer files
- [ ] Remove your email from the admins setting
- [ ] Remove your username from the db config

Those last two are just because it won't work on the live server with those settings. I also want to clean up the code so it conforms to some python styling guidelines.
Thank you.
(Relates to #8 )

zaneadix:
Nice :). Thanks for responding so quickly, Paul. I'll clean the config and squash and see if I can get that to you tonight.

Is there any testing that can be written for the models or will most of that just need to happen on the API calls?

phalt:
@zaneadix Testing for the correct / expected attributes would be useful. Try and create a Egg / Move / Description and then populate it's fields. Make sure you can access all of it's attributes. I see you've used a lot of Mixins. I personally try to avoid them but that isn't a problem as this has made the models quite composable.

zaneadix:
Migrations have been squashed and settings have been cleaned up.

I got most of the way through all of the Move models last night but didn't get all of them in the build script. I'll have them soon. I'll also look into some tests for the models.

Anything else I should do on the pull?

phalt:
Migrations are now missing ;)

phalt:
For now I will leave this open until I want to build and review it. This is only because I don't want to merge it and then forget about it. Having the PR open means I'll be aware of it. I might find some spare time this weekend. It's looking good so far. Thank you!

zaneadix:
Migrations are back...

Sounds good!

phalt:
Right, I'm going to try merge this today and make sure it works nicely. I can then start trying to build the API front end part on top of it. @zaneadix, could you look into building the Pokemon and Move models too?

phalt:
Just two things @zaneadix: can you write a quick how-to on how to load the data using the scripts you've built? Put it in the build.py file. Second thing: Will it load into a dev env using sqlite?

zaneadix:
@phalt I actually have done most of the work on moves and pokemon. They just need to be added to the build script. I actually did this against full on Postgres. My bad. I'll have to try and load them into SQLite tonight. Ilove get a how to written tonight as well. Maybe I should also add something to the reader?

zaneadix:
Readme*

phalt:
README is a great place for it

PR #35, opened 2015-04-13
Completed Models. Build script improvements.

zaneadix:
Wanted to get this out yesterday but it just took longer than I thought it would. Below is a list of all that I've added. This should address anything you've mentioned up to now. I'm gonna look to add Berries, Items and Regions from veekuns stuff.

-Added models for Pokemon, Moves, and all other remaining models.

-Added new models to build script

-Build script will now clean and reset auto-incrementers for each table
in pokeapi_co_db

-Build script now runs against both postgres and sqlite (dev)

-Some v2 cvs files had no data in them. Pulled in the data from veekun
where necessary.

-Added all models to admin site registry

-Added instructions for building out db in README


zaneadix:
Added every last bit of data (Except for that Conquest stuff) and models for each.

Have you had a chance to work against the data at all? Have you run into any issues?

phalt:
This has become a huge pull request. I'm not going to merge it at it's current size. I know it has all the pokedex csv data and that is why. However, could you separate out your code from the build scripts? I can't review it this way.

zaneadix:
Yep. Sorry I didn't even really consider that. I'll see if I can split them out some time today

zaneadix:
Closed to separate data additions and code

PR #39, opened 2015-04-21
New data

zaneadix:
Just copied over more data from veekun. 
With this commited the only data files left over in veekuns repo would be for Pokemon Conquest.

Data and models added for:
Pokemon
Moves
Pal Park
Berries
Regions
Pokeathlon
Contests
SuperContests
Items
Encounters
Locations


matthewdias:
I'm guessing this doesn't have ORAS data, since veekun hasn't been updated since the XY changes?

phalt:
@zaneadix is this based off your other PR? Does that need to be merged and rebased first?

zaneadix:
@phalt Yeah it would probably be best if they were merged first. I really only separated them out so they were easier to review.

@matthewdias Yeah this doesn't have ORAS data. Maybe we should consider jumping in on Veekun's stuff.

phalt:
@zaneadix I am merging this :) Not deploying it yet though

PR #55, opened 2015-09-27
Build fix

zaneadix:
There was some an issue renaming a field in the generated migrations. Made an edit to fix.


phalt:
Thanks for this @zaneadix

PR #56, opened 2015-10-02
Update v2 db setup instructions to use local config

lazerwalker:
Trying to set up a local instance, I ran into some minor trouble trying to import the v2 since the README's command to open a python shell wasn't passing in the settings config file flag, resulting in the v2 db import script trying to hit a postgres instance instead of the sqlite file.

This just adds that flag to the README.


zaneadix:
Thanks @lazerwalker

PR #58, opened 2015-10-03
Building v1 data

zaneadix:
Tried to build out the v1 data just to help out people trying to start contributing. I made some minor changes and added some notes to the README. 

I ran into an issue though with Egg Groups. I noticed there isn't even a csv for egg_groups data/v1, only for pokemon egg groups. This is keeping me from populating all the tables. Do you know why this stuff isnt in there?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,14 +42,31 @@ If you ever need to wipe the database use this command:
 $ make wipe_db
 ```
 
-## V2 Database setup
+## V1 Database setup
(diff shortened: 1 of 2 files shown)


phalt:
@zaneadix Hey, I'm not sure why the data isn't there. I might have added it manually? This looks good anyway. if you're sure it works go ahead and merge :)

PR #62, opened 2015-10-18
Tests done for current api endpoints

zaneadix:
Just figured I'd make this visible to see if anyone has any thoughts on whether this is an alright way to write the tests. Trying to abstract the different data requirements similar to the way the models are handled.


phalt:
@zaneadix This is looking good. Looks like you've had to make a few code changes to get it to work too? Nice to see the tests finding bugs :grin:

zaneadix:
@phalt yeah haha the tests definitely helped spot some inconsistencies in the way some data was served. 

I've run into an issue trying to verify urls though. On the hyperlinked resources I want to be able to check the urls but cant find any way of building out the url dynamically based on the environment without access to a request object. I don't want to hard code the host and port anywhere so I'm thinking maybe theres a way to have it set in the settings file for access throughout the code. Do you have any idea what would be the best route to take?

zaneadix:
Along with the api tests a number of changes were made for project consistency.

Veekuns csv files don't really have a very consistent naming convention so when I was making models I guess I got a little confused. Some files that were called *_prose.csv might contain names, descriptions, flavor text, etc. After looking at things I thought it best to create an *EffectText model for some data sets and scratched their origin *Description models (because they didn't contain descriptions). 

Some models were still pointing to integer values rather than foreign key relations so those should all be taken care of now.

Also fixed some inconsistencies in serializer behavoire/relationships.

If things look good to you @phalt  I think this should be good to merge in

phalt:
Amaze :+1:

pokeapi-machine-user:
The updater script finished its job and the generated data didn't change. The deploy was thus skipped. For further information check [CircleCI's builds](https://app.circleci.com/pipelines/github/PokeAPI/pokeapi) and [logs](https://circleci.com/gh/PokeAPI/pokeapi/704).

PR #63, opened 2015-10-23
Include support for all formes/pokemon or ability to retrieve picture

xplustwo:
Suggest adding support for all different formes, as well as any new pokemon.
If possible, also add support to retrieve the picture in case of graphical implementation.


zaneadix:
Sorry but this needs to be submitted as an issue rather than a pull request. As far as I can tell you haven't done any work toward your suggestion.

phalt:
@xplustwo thanks for the work but as @zaneadix pointed out you should make an issue first and then work on a pull request. Also, it appears like you need to rebase!

PR #70, opened 2015-11-12
Add a Dockerfile to build the application as a container

OrangeCrush:
Wanted to share the Dockerfile I have been using with the group.  Tested and working well.  Building the container takes about 20 hours (The database is rebuilt each time)

Open to any feedback / suggestions.  If we can get the build time down (commit a prebuilt DB into master?) we would be able to set up an automated build pipeline to dockerhub so people could pull down and use the app without the need to build it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Build the app on top of Ubuntu
+from ubuntu
+
+# Patch and Install Dependencies
+RUN apt-get -y update && apt-get -y upgrade
+RUN apt-get -y install git python-dev make python-pip libpq-dev
+
+# Add the application code to the image
+ADD . /root/
+
+# Set a working directory
+WORKDIR /root/
+
+# Build the application
+RUN make install
+RUN make setup
+
+# Build the v2 database
+RUN printf "execfile('data/v2/build.py')" | python manage.py shell --settings=config.local
+
+# Expose the app and serve the API.
+EXPOSE 8000
+CMD python manage.py runserver --settings=config.local 0.0.0.0:8000
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ WORKDIR /root/
 
 # Build the application
 RUN make install
-RUN make setup
+RUN python manage.py migrate --settings=config.local
 
 # Build the v2 database
 RUN printf "execfile('data/v2/build.py')" | python manage.py shell --settings=config.local
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,21 +3,29 @@ from ubuntu
 
 # Patch and Install Dependencies
 RUN apt-get -y update && apt-get -y upgrade
-RUN apt-get -y install git python-dev make python-pip libpq-dev
+RUN apt-get -y install git python-dev make python-pip libpq-dev 
+RUN apt-get -y install postgresql postgresql-contrib
+
+CMD mkdir -p /app && chown -R postgres:postgres /app
 
 # Add the application code to the image
-ADD . /root/
+ADD . /app/
 
 # Set a working directory
-WORKDIR /root/
+WORKDIR /app/
 
 # Build the application
-RUN make install
-RUN python manage.py migrate --settings=config.local
+RUN pip install -r requirements.txt --upgrade
 
-# Build the v2 database
-RUN printf "execfile('data/v2/build.py')" | python manage.py shell --settings=config.local
+# Start postgres database and use it while it is running in the container
+# Create the default db user (ash) 
+RUN sudo -u postgres service postgresql start                                 && \
+    sudo -u postgres psql --command "CREATE USER ash WITH PASSWORD 'pokemon'" && \
+    sudo -u postgres createdb -O ash pokeapi                                  && \
+    python manage.py migrate --settings=config.docker                         && \
+    printf "from data.v2.build import build_all; build_all()" | python manage.py shell --settings=config.docker
 
 # Expose the app and serve the API.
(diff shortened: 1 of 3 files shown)


phalt:
twenty hours? Is that a typo?

phalt:
Can we add an option to populate minimal test data. 20 hour build times is not practical. The scripts can be run in 20-30 minutes so I don't understand why a build would take that long.

OrangeCrush:
Anytime I build the complete database per the README it takes close to 20 hours for the following command to complete-

printf "execfile('data/v2/build.py')" | python manage.py shell --settings=config.local

I will see if I can decouple the database from the container into a "data" container so that the app can be rebuilt separately from the database.

Am I do something wrong here?  I do not know Django very well... Let me know.

zaneadix:
Woof. 20 hours sounds crazy. Maybe this has something to do with local settings? I'm honestly rarely working against the lite db. Do you have a full postgres instance running? Have you tried it without running local settings? Regardless that's a problem. It's usually taking about 20 minutes for me right now.

OrangeCrush:
OK.  So I have had a change to get back to this.. I have added 2 new commits to the PR.

I did take a look at the postgres setup and it did reduce the build time by quite a bit for all of the "v2" data.  It now builds in roughly 90 minutes on my desktop.. Thanks for the suggestion

```
Successfully built 3fbd129a5aeb

real    92m21.323s
user    0m0.302s
sys     0m0.112s
```

Unfortunately this is still too long to set up an automated build on dockerhub for right now. (see the issue I have open [here](https://github.com/docker/hub-feedback/issues/512) )

You can check out the container image I have pushed into dockerhub [here](https://hub.docker.com/r/orangecrush/pokeapi/).

What do you guys think about merging this?  Let me know your thoughts.  
Thanks

phalt:
Hehe I think this is a problem we're not going to be able to avoid easily. It's a *lot* of data after all. Perhaps we can make the docker file just build the framework then we can manually run the data afterwards? Or perhaps we can just load some test data (like 1 or 2 pokemon and some sample data for the rest)?

zaneadix:
It'd be great to have some decently fleshed out test data. Work done in #75 which has already been merged in has taken care of a lot of what makes this a problem by breaking out the data building into chunks so you don't have to build the whole thing at once.

OrangeCrush:
I'm curious if there is any interest in merging this in as is at this point?  We could have a separate container with test data applied if someone has use cases.

I think there are a lot of benefits of using the full database in a docker container with this project... 

* *Super* useful for the end user to be able to run the latest API code locally with one command.
* For the heads down pokeapi dev's it's useful to be able to quickly test if the database still builds from scratch
* Drastically improve any kind of deployment and minimize downtime.

phalt:
@OrangeCrush Happy to merge it but I have two questions:

1) How dramatically will this affect the current production project? I am not 100% familiar with docker and I want to avoid downtime getting this up and running.
2) Will this reduce barrier to entry for people who don't have docker. I.e. - can I still dev normally without it?

OrangeCrush:
@phalt -

1) The changes I made give people the option to build the app as a "Docker Container".  Nothing needs to change for the production project, unless you are interested in using docker to deploy it.  Using docker would be completely optional.  Most projects are not using Docker in production.

2) Again, no one is required to build the application as a container, so this would not have any impact on those who may not have access to use docker.

Speaking of "barrier to entry" though, IMO this drastically reduces the barrier for other folks who would like to consume the API for their own projects.  Being able to build (or download) the entire running application to a desktop environment make's it much easier to do development.

Ultimately what docker allows you to do is programmatically build (and run) an application in a brand-new, fresh environment.  Think of the container as a service running on your machine.

I can definitely take some time to update the documentation if you think that is needed before merge.

I'm curious if you are developing on Linux yourself?

nimaje:
a thought:
How about a repo for configs which aren't needed, but could be useful, like this Dockerfile,
because there are more programms which do similar tasks and why should we only support some of them if we add this others could be requested with  reference to this issue.
A repo for such configs would sepperate optional configs from the main project. A weaker separation would be a folder `optional_configs` (or similar).
As far as I know docker only works on Linux and I don't know how it would use `apt` on something like arch or gentoo.

OrangeCrush:
@nimaje -
Docker can be run on any 64 bit operating system (including Mac and Windows).  I personally find it easiest to use Linux (for everything =] ) which is why I was asking about it.

Since the Dockerfile I wrote uses Ubuntu as it's parent container, the "apt" binary is already packaged inside of the container's chroot.  Although I have not specifically tested this container on gentoo or arch linux.. the entire crux of Docker is to solve this problem (build once, deploy anywhere), so I am more than confident that it will run anywhere without issues.  I have personally tested the container on Centos v6 & v7, Ubuntu v14 & v15, and Atomic v1.6.

Finally, I don't think it makes sense to keep the Dockerfile in a separate git repository.  Building the app in a container is dependent on the source code, so in my opinion they should live together.

OrangeCrush:
I think there may have been some confusion with something I did not realize was in this PR.  There are no changes to the default "config/local.py" file.

I amended my PR to appropriately reflect this... I hope this clears things up as well

phalt:
@OrangeCrush perfect answers :) I dev on mac os x with docker, just not for this project as it was before docker existed :D

nimaje:
I suggested a repo as *seperation*. I know that Docker *will need* the source if you use it,
but this project *doesn't need* the Dockerfile ( it could be useful for some).
As a weaker seperation I suggested a subdir which indicates everything in it is optinal.
(Just to explain my thought)

PR #73, opened 2015-11-30
Docs

zaneadix:
Started the documentation effort. Just writing everything out in a README within the pokemon_v2 directory for now. Currently some things have been changed that should not have been. I will work this out.


config/settings.py
```diff
@@ -89,7 +89,7 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'pokeapi_co_db',
```
phalt, line 92: This needs to remain as `pokeapi_co_db` for the live site. Perhaps a debug setting?

phalt:
This is looking good. I would add in an example request and response for each resource if you could, one that can be copy and pasted :+1:

zaneadix:
https://github.com/zaneadix/pokeapi/tree/docs/pokemon_v2

@phalt I think this should be good. I added example responses with all arrays reduced to a single index to keep things as short as possible. Any thoughts on what else could be added?

pokemon_v2/serializers.py
```diff
@@ -1036,6 +1037,28 @@ def get_ability_pokemon(self, obj):
 
         return pokemon
 
+    # def get_flavor_text(self, obj):
```
phalt: Is this supposed to be part of this PR?

zaneadix: Whoops nope. This was an experiment that seemed not necessary in the end. Thanks for catching. Took it out.

PR #75, opened 2015-12-05
Improve db build

nimaje:
The db build module was a huge mess with only two functions and the main code on module-level. I have changed that. Now you can import it and only call the functions you want to execute.
Furthermore I have made some little changes to make it python3 compatible.
The db build has needed 22h on my computer (local settings).  All test passed, but I don't know if there are tests about the completeness of the db.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,10 +68,11 @@ $ python manage.py shell --settings=config.local
 ```
 run the build script with
 ```
-$ execfile('data/v2/build.py')
+$ import data.v2.build
+$ data.v2.build.build_all()
 ```
 Each time the build script is run it will iterate over each table in the database, wipe it and rewrite each row using the data found in data/v2/csv.
-If you don't need all of the data just go into data/v2/build.py and comment out everything but what you need to build the tables you're looking for. This might be useful because some of the csv files are massive (pokemon_moves expecially) and it can take about 30 minutes to build everything.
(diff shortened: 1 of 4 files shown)


zaneadix:
Thanks @nimaje. Gave this a quick glance and it looks awesome. This project is my first time coding in python so i've probably done a bunch of questionable stuff. I'll pull this down and check it out soon. Would you also be able to update the readme? Is there still a method that would allow us to run all methods?

nimaje:
yes. The methode in the readme should work, but I don't know django (`__name__` has to be `__main__` in the module that it run) . To run all methods there is `build_all`.

cstrachan88:
I noticed there was a duplicate call to PokemonGameIndex (at line 2048) so I edited your updated build .py for v2, @nimaje. I haven't used github in a while so I don't know if there's anything else I need to do besides fork and edit the file.

nimaje:
thanks @cstrachan88. such things were meant with mess (I think this was the only dublicate). And the use of `enumerate` to skip the header of the csv (not changed yet).

zaneadix:
I've tested everything out and this seems to be good to go. I've made a small pull request with some documentation changes. Can you merge that in and then I'll merge this into the main branch. Thanks :)

nimaje:
I have made some changes because of the starimports in the examples and the list with the build functions was under the v1 section I have move it under the v2 section

zaneadix:
whoops. Thats what I get from trying to sneak edit this stuff at work. Thanks @nimaje nimaje

PR #78, opened 2015-12-27
remove templating from strings when building database

zaneadix:
Cleaned up effect text strings.
Example:
        []{move:dragon-tail} will effect the opponents [HP]{mechanic:hp}.
    Becomes:
        dragon tail will effect the opponents HP.

### DIFF
--- a/data/v2/build.py
+++ b/data/v2/build.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 import csv
 import os
 import os.path
+import re
 from django.db import migrations, connection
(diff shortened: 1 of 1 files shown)


phalt:
Nice, will I have to run the db build script again after this @zaneadix ?

zaneadix:
Yeah at the very least the build methods for abilities, items and moves but it'd probably just safer to run build_all()

phalt:
@zaneadix Okay, I can merge this and then run it again this week :)

PR #79, opened 2015-12-28
updated data with all of latest veekun updates

zaneadix:


phalt:
Ah interesting, how likely is it that something like this will have to be updated?

zaneadix:
Do you mean how likely is it that we would want to update data with the most recent veekun changes? I don't think we would need to do it often. I was just planning on keeping an eye out for anything significant or when new games are released. I just did this because I figured it would be good to launch with everything totally up to date when we're ready.

nimaje:
Do we use the data from veekun without changes (for processing with the build-scripts)?

If yes, why should we have a copy of this data. We could write a script to download the most recent version and than run the build-script on these files.
There would be no probleme with being up to date with the veekun data. You only have to rerun these scripts when there are changes.
These downloads could be big, but why should the data be in this repo if it has no changes to the data from veekun.
As well images why should these in this repo **if** we have a **good source** for the images. These files (veekun's csv and the images) are only needed by the server not by the people editing the code. Why not downloading these when needed.

If there are changes in the csv files or no good source for the images why not make a repo for the data?

zaneadix:
There are no changes to the csv files. I think this is a fine idea if you want to tackle it. Downloads wouldn't be big. They're just csv files. I actually have a repo set up locally that pulls only those files from veekun's db. As for the images, I personally don't have much opinion one way or the other. I imagine when @phalt started this project his original intention wasn't to invite the open source community in so when he put it up here, he just put it up here.

nimaje:
The csv files are in the repo [veekun/pokedex](https://github.com/veekun/pokedex) @ `pokedex/data/csv/`. The buildscript could be writen that you can give it the directory of the csv files (*If* there is a local copy why redownload. You only have to remember updating the repo copy) and a script to download the files out of the repo into the normal directory.
There would be a little change in the db-build script for those who have/want a local copy of veekuns pokedex and a script to download these files for those who don't want a repo copy. I have to read a little about the structure of github. I think this could be easy to write, but I don't know.
Veekun has some images (http://veekun.com/dex/downloads) but I don't know were they are from and they are only up to gen 5.

phalt:
It's easy enough to update the data we have here when ever there are significant changes (new release of games usually mean slight stat changes for some pokemon, and new pokemon). I can rebuild the data whenever that happens. :)

zaneadix:
@phalt just want to call attention to this one and see if we're in agreement on merging it in. Any reason why we would want to hold off on it?

phalt:
@zaneadix nope, just had a busy weekend :)

PR #81, opened 2016-01-02
New ui

zaneadix:
@phalt I'd like to get your thoughts on this. I've done some work to set up a new ui. I've written it in angular (fastest way I know to work). Right now what I've done mirrors v1 pretty closely but there are some things I want to play with. 

I've set up the urls so that users can still navigate to and through the v1 ui. I'm trying to get this knocked out ASAP honestly just because it would be a lot of work to create the docs page for the v1 site and I figured it would make more sense to just work on everything.


nimaje:
I am against a new ui.

I think this project is about a good api for information about pokemon.
The homepage should only need some (mostly) statical htmlpages to inform about the project and the documentation.

I haven't looked much on your changes but have you build a js-app for the data?
If yes, why this would be another project which could be hosted on the same server if @phalt thinks this is needed, but not something for this project.

Moving the configs in the proper django-app was a good idea.

What is this gulp files? This looks like some preprocessing.

And why this  minimized versions of the files they should be build for the server and on the vcs the normal versions should be used (Who would want to edit minimized files?).

For what is this ``package.json`` file?

zaneadix:
@nimaje The new ui is still pretty much what it is currently (functionality-wise) but I thought it could use some character (shoot me I'm partial to ui/ux). I personally think giving the site a more personalized feel will attract more people to use it. I've also added a page to list contributors (so you'd be more likely to be recognized for your help :) ). Maybe I'm alone in this. If so I can drop it.

I still think it's about a good api as well as that's where the majority of the work has gone. Gulp is a task runner for JS.  It's just handling minifying js and sass. Checkout pokemon_v2/client for the angular ui code (you wouldn't make changes to minified files). package.json is for npm dependencies (which is where gulp runs). It's all pretty standard tooling among a lot of ui projects.

nimaje:
@zaneadix These js scripts look like you provide mostly statical sites with js. Why you don't use django and templates for this?
Use preprocessing to easier write this html and css files is a good idea, but I see no need for client-side js. Only use js on the client side if it is needed else browsers like lynx can't use the website.
And why have you not included these minified files in ``.gitignore``?
Who wants autogenerated files in a versioncontrolsystem when the sources are already there?
The contributors are in ``CONTRIBUTORS.txt`` a page could  be generated out of this file.

zaneadix:
@nimaje should we really be worried about whether or not Lynx can read the site? It seems pretty moot to me. The main reason I didn't use django templates is because I'm not a python or django developer really which I'm sure is evident in places. I'm much more efficient in js and I wanted to get something done quick but well. I really don't see a huge disadvantage one way or the other. Single page sites are a very normal thing. The only reason I didn't add the mines files to the hit ignore is because this is a work in progress and I just wanted people to be able to pull it down and view the up without running anything. I wouldn't push it to the repo the way it is currently. I was just using the github api to get the contributors. I don't know anything about using the contributors file.

nimaje:
lynx was a extreme example, but I have js disabled in my browser, because websites load faster without js and most times js isn't needed (and I see almost no ads).
I don't like sites which aren't useable without js, but actually don't need js.
I think things which could run on the webserver should run on the webserver.
OK I can see your point why you have not used django templates, but most of these js files look like some boilerplate for serving a statical site.
I django this should be
```python
urlpatterns = [
…
url(regex, TemplateView.as_view(template_name=name)),
…
]
```
This should work on normal html files as well as on django templated files.
For the contributors there could be something like
```python
CONTRIBUTORS =  # some github api call or read CONTRIBUTORS.txt
…
def generate_contributors_page(request):
    return render(request, 'pokeapi/contributors.html',
                  {'contributors': CONTRIBUTORS})
…
urlpatterns = [
…
url(r'^contributors$', generate_contributors_page),
…
]
```
And as easy template 
```
…
    {% for contributor in contributors %}
        {{ contributor }}<br>
    {% endfor %}
…
```
This `CONTRIBUTORS.txt` file is on the project root it is a simple text file created by phalt nothing special.

phalt:
I am also partial to UI/UX but I don't think Pokéapi needs a face lift. I think angular over complicates what is essentially just documentation and a simple little ajax call on the home page.

I **do** think we can rebuild the documentation to work by parsing markdown. This will make it far easier to update the documentation in the future.

I feel bad but - can we remove the angular and use a simple markdown parser for the docs? Checkout the Star Wars API project I made - it does exactly what we want: https://github.com/phalt/swapi/blob/master/swapi/templates/documentation.html#L44L47

zaneadix:
@phalt should v1 and v2 both count as calls to the same api?  Or would we want to keep those separate? Also for object counts (like on the about page). I've also been trying to think of a logical way to group the endpoints so that navigation on the docs page can be simplified. Something like below. Does this make sense?

Berries
- Berries
- Firmnesses
- Flavors

Contests
- Contest Types
- Contest Effects
- Super Contest Effects

Items
- Items
- Categories
- Attributes
- Fling Effects
- Pockets

World
- Habitats
- Locations
- Location Areas
- Pal Park Areas

Pokedexes

Pokemon
- Colors
- Egg Groups
- Evolution Chains
- Forms
- Habitats
- Pokemon
- Shapes
- Species

phalt:
@zaneadix Small UI tweaks are fine. We should track the API usages separately, but we can keep a "total" too for the front page.

As for the question about counts - let's move that to another issue so we're not talking about multiple things :laughing:

zaneadix:
@phalt This isn;t going to be done as soon as I'd like. The implementation of markdown_deux even with extras applied is clashing with quite a few things in docs.md. Searching for answers.

zaneadix:
@phalt Alright so I did quite a bit of changes but really just simple ones. I updates the ui (biggest difference is probably the header block). I made a few changes to the v1 docs. I think I mostly just moved the parameters into tables so the view was similar to what I had done with the v2 markdown docs. I also made some updates to the sidenav so that things would grow and shrink in a useful way on the v2 docs. Documentation has one navigation tab and has a v1/v2 toggle in the sidenav. 

I made some changes to the about page. Biggest thing is probably the resource counts which are now just focused on V2 resources. 

The hits ResourceView model has been updated to add a "version" column which defaults to 1 so it should be easy to migrate in the deployed environment. The total_count method still returns a count of all api calls (v1 or v1) but you can also call it with a number to get specific version counts.

I hope everything looks alright.

phalt:
@zaneadix so to understand correctly: this no longer tracks V1 resource hits?

zaneadix:
No it does. It tracks both. The count on the front page is a total of both v1 and v2 resources. I just made some changes so that would could track separately if we wanted to. Sorry for any confusion.

zaneadix:
@phalt Do you have any other questions? I hope I didn't make too drastic of changes.

nimaje:
@zaneadix where do we use a frame with the name `none`?
You are using in `templates/about.html` the html a tag with `target="none"` what means the link should be opened in an frame with the name `none` but I don't see where you make use of frames, where should the link be opened?

zaneadix:
Probably should have changed it to target="_blank" but none seems to work
the same. It's just to open the link in a new tab or window

phalt:
@zaneadix no it looks good. There isn't a brilliant way for me to test or look at this. I'll try and merge it over the weekend and get it on beta.pokeapi.co :)

zaneadix:
@phalt Im not quite sure what happened but I was looking over the code in the pull request earlier today and noticed that this isn't in the completed state I thought it was. I may have just pushed from the wrong computer without knowing it but this is currently missing some things. I'm working on fixing this right now.

zaneadix:
@phalt ok so I apparently had stashed changes for some reason right before I committed and pushed. Im really not even sure how or why. That's in the past now. I've pushed up what should have been committed. (Also the number of files changed was cut in half. I should have seen it as a sign that something was wrong when it was in the 90's)

Issue #85, opened 2016-02-01
Identical Move Records Getting Sent Multiple Times in Summary


SiMylo:
I've noticed that if a move has different learn levels in different version groups, the exact same move record will be copied multiple times when listing a pokemon's learnset.  This appears to be because the query sorts by 'level' before looking for distinct 'move_id' values.  According to Django documentation, this will happen because internally it's keeping track of the level so each entry looks different.  I have a fix for this locally, but I have no idea how to testing.  Should I submit a pull request, or can someone point me to a decent primer on how our testing framework works?

phalt:
@SiMylo Hey, thanks for raising this.

Is this for V1 or V2?

zaneadix:
From the description it sounds like it would be V2. @SiMylo Would you be able to post an example of the bug here? I'm just interested :) As far as tests go you could checkout tests.py in pokemon_v2 and look the `setup_pokemon_move_data` method and the `test_pokemon_api` method. You could most likely test to make sure your fix works just by adding a few moves with the same name, different version group and different learn level and then assert the results.

SiMylo:
@phalt this is for V2.  I'll be more specific next time.

@zaneadix If you search for "thunder-wave" in the following file you can see 5 copies get returned from api/v2/pokemon/25/ .  I will check out those files and see if I can get it tested and then submit a pull request.
[api.v2.pokemon.25.txt](https://github.com/phalt/pokeapi/files/114263/api.v2.pokemon.25.txt)

zaneadix:
Sure enough. Nice catch @SiMylo. Thanks for the taking this on :)

zaneadix:
#88 takes care of this. Thanks @SiMylo!

PR #86, opened 2016-02-03
Fix for Issue #85 (Duplicated move records)

SiMylo:
Changes our query to order by the same column that it is using to determine if objects are unique.  Also added testing.


SiMylo:
Sorry for the mess, it looks like it put all of my commits here.  I'm quite new to git, so I had some trouble when I tried to merge in the latest version.
2109213 was my original local fix.
648d933 was a failed attempt to merge phalt/master to my repo. 
8956615 was an attempt to revert but it reverted to the wrong version.
a36bdd0 was my successful attempt to get my repo in sync with phalt/master .
54b174f contains all of the relevant info.

If someone has a suggestion on how I can cancel this request and get a cleaner one, please let me know, because I agree this is a mess.

PR #88, opened 2016-02-03
Fixing Issue #85 (Duplicated Move Data)

SiMylo:
Contains a fix and testing for the following issue:

I found that we are sometimes sending multiple exact duplicates of
pokemon move data when there are different levels at which a pokemon
learns a move for different version groups.  I found that the problem
was that we were using order_by('level') and duplicate('move_id') .
This link explains why this behavior happens.
https://docs.djangoproject.com/en/1.7/ref/models/querysets/#distinct

### DIFF
--- a/pokemon_v2/tests.py
+++ b/pokemon_v2/tests.py
@@ -1806,16 +1806,10 @@ class APIData():
         return pokemon_item
 
     @classmethod
-    def setup_pokemon_move_data(self, pokemon, level=0, order=1):
-
-        move = self.setup_move_data(
-            name='mv for pkmn')
+    def setup_pokemon_move_data(self, pokemon, move, version_group, level=0, order=1):
 
         move_learn_method = self.setup_move_learn_method_data(
             name='mv lrn mthd for pkmn')
-
-        version_group = self.setup_version_group_data(
-            name='ver grp for pkmn')
 
         pokemon_move = PokemonMove.objects.create (
             pokemon = pokemon,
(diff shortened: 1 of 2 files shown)


pokemon_v2/tests.py
```diff
@@ -3581,13 +3581,18 @@ def test_pokemon_api(self):
...
-        self.assertEqual(response.data['moves'][0]['version_group_details'][0]['move_learn_method']['name'], pokemon_move.move_learn_method.name)
-        self.assertEqual(response.data['moves'][0]['version_group_details'][0]['move_learn_method']['url'], '{}{}/move-learn-method/{}/'.format(test_host, api_v2, pokemon_move.move_learn_method.pk))
+        #  Make sure that we only got one move back, but that we got all of the distinct version group
+        #  and learn level values. (See issue #85)
+        self.assertEqual(len(response.data['moves']),1)
+        self.assertEqual(response.data['moves'][0]['move']['name'], pokemon_moves[0].move.name)
+        self.assertEqual(response.data['moves'][0]['move']['url'], '{}{}/move/{}/'.format(test_host, api_v2, pokemon_moves[0].move.pk))
+        self.assertEqual(len(response.data['moves'][0]['version_group_details']),len(pokemon_moves));
+        for version_group in range(0,len(pokemon_moves)):
+            self.assertEqual(response.data['moves'][0]['version_group_details'][version_group]['level_learned_at'], pokemon_moves[version_group].level)
```
phalt: Could you reduce the line length please?

Perhaps something like 

```
expected = pokemon_move[version_group].level
actual = response.data['moves'][0]['version_group_details'][version_group]['level_learned_at']
self.assertEqual(expected, actual)
```

It's easier to read :)

Tests shouldn't try to do anything complicated!

Thank you :D

zaneadix:
@phalt @SiMylo This is really something I should have been doing the whole time as well. Derp.

SiMylo:
@phalt I can keep shortening.  Without being too familiar with the code base, I'm not sure what the desired balance between line length and number of lines is.

phalt:
@SiMylo where possible, new code should try and follow the PEP8 standard for Python :) https://www.python.org/dev/peps/pep-0008/

Older code isn't, so we'll fix it as we encounter it!

SiMylo:
Noting that I did upload changes.  I think this request has fallen to the back burner.

phalt:
@SiMylo sorry for the slow responses - got a day job to do :laughing:

Nice changes, and the code looks nice! I'll :+1: it now but let @zaneadix approve too as he worked on parts of this code.

SiMylo:
@phalt No problem.  I wasn't sure if there was any way for you to know that there was an update here without posting a message.  Glad it looks good.

zaneadix:
@SiMylo Sorry for the slow response. Pulled it down and checked it out. Change is good. Thanks for the good work :)

Issue #107, opened 2016-02-24
Access to a pokemon's pokedex description


fmonniot:
Hi there !
First of all, thank you for this API!

I would want to know how I can get a pokemon description (as in the pokedex).
I have seen that `data/v2/csv/pokemon_species_flavor_text.csv` seems to have the data I look for but I can't find how to get _via_ the API.
Any idea ?

Thanks

phalt:
@fmonniot have you tried the V1 API for this?

fmonniot:
Well maybe I could (I haven't tried) but I would prefer not to use two
version of the API of I can.
I'm ready to open a PR if this is not currently implemented in V2 and if
needed

phalt:
@fmonniot Let me ask @zaneadix if this is something we've merged into the project but not shipped yet :)

I need to get myself a weekend free so I can ship some recents changes :)

nimaje:
At [data/v2/build.py line 1909 - 1918](https://github.com/phalt/pokeapi/blob/master/data/v2/build.py#L1909) the data goes in the db as `PokemonSpeciesFlavorText`.
I don't know if this is accessible trought the v2 api. I didn't find anything in the docs neighter online nor in the repo.

zaneadix:
@phalt @fmonniot @nimaje So strange. I sincerely have no idea how I would have missed this. Working on implementing this right now.

fmonniot:
@zaneadix No problem, I guess you are more interested in pokemon's statistics than in their description :)

phalt:
Resolved in #111 - will let you know when it is shipped @fmonniot, then I can close this :)

vihanggodbole:
Hey! Just a quick question - Any ETA on when will this be shipped?

phalt:
@vihanggodbole my Sunday will be taken up mostly by shipping new things.

vihanggodbole:
Since this is closed but I still can't find any description when I call the api on any pokemon, could you tell me how to access the description? Thanks!

fleeting:
@vihanggodbole Thanks for that! I was also looking for the pokedex description with no luck until seeing your comment.

zaneadix:
@fleeting @vihanggodbole You're looking for the `flavor_text_entries` param on the pokemon-species resource (NOT the pokemon resource). I'm seeing them pop up now thanks to @phalt 's work over the weekend :)

fleeting:
[...]

PR #111, opened 2016-02-25
Flavor texts

zaneadix:
Add in the flavor text entries on pokemon species that should have been there in the first place.


zaneadix:
@phalt should fix #107

phalt:
Awesomes, nice one @zaneadix !

PR #124, opened 2016-03-05
Version bumps, removing whitespaces

phalt:
Testing out circleci and fixing some things


phalt:
YAY it works. @zaneadix We now have automatic builds and test suites in pull requests :)

PR #136, opened 2016-03-09
Updated About Page

LawlietBlack:
I hope you don't mind, I refactored a lot of the code on views.py. The numbers may appear quite different, and that's because the old line count was actually the object count, where the object count was just counting a subset of the objects, possibly the objects from v1. There was no actual line count before, as far as I could tell. The new line count and objects counts are accurate and cover both v1 and v2, since the API serves both. All of the list items under "How much information is stored here?" are from v2 data.

I also corrected a couple of file references that were outdated.

### DIFF
--- a/config/views.py
+++ b/config/views.py
@@ -1,7 +1,6 @@
 #!/usr/bin/python2.7
 # -*- coding: utf-8 -*-
 import inspect
-import os
 import re
 
 from django.shortcuts import render_to_response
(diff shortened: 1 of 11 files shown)


phalt:
@LawlietBlack thank for you this contribution, it's great that you've tried to reduce the amount of work!

It appears that we're opening up a file and reading it's contents, and this will happen each time the view is loaded? IO for a file is very, very slow and I feel it would be better to just store the values hard coded if this is the only way to get a true representation of the amount of data we store. A secondary note, this data could also be inconsistent with what is actually available in the database, as we load the content from these csvs manually.

_However_, this is really good progress, and I think it's going in the right direction :)

LawlietBlack:
Maybe it's a semantics issue, what I understood as "built from _ lines of data" was referencing the CSV files. I agree that hard-coding a value would probably be better in that case, though... I can replace "built from ___ lines of data" with "built from nearly 1,000,000 lines of data"

LawlietBlack:
updated commit with those changes

LawlietBlack:
Was there anything else you wanted me to change?

config/views.py
```diff
@@ -17,202 +27,21 @@ def _total_site_data():
...
+    # v2 objects count
+    for name, obj in inspect.getmembers(models):
+        if inspect.isclass(obj) and hasattr(obj, 'objects'):
+            data[convert(name)] = obj.objects.count()
+            data['total_resources'] += obj.objects.count()
 
-    data['total_lines'] = lines
-    data['total_resources'] = resources
+    # v1 objects count
+    for name, obj in inspect.getmembers(models_v1):
```
phalt, line 42: Using `inspect` feels like a round-about way to solve this problem. Why did you choose `inspect`?

PR #162, opened 2016-04-17
Optimize data loading time by 76%

mjhayes412:
Still learning this process but hopefully this might help in development....

Here's how the improvement works:

1. Build project as is with build_all(); //took me 26m 7362s
2. Once all data is loaded, run dump_all(); //took me 49.7914s
3. Anytime afterwards, if 'make wipe_db' is executed, after 'make setup' the following command will build the entire database 76% faster:

python manage.py loaddata pokemon_v2/fixtures/*.json --settings=config.local

//took me 6m 4141s

### DIFF
--- a/data/v2/build.py
+++ b/data/v2/build.py
@@ -23,6 +23,8 @@ import re
 import json
 from django.db import connection
 from pokemon_v2.models import *  # NOQA
+from django.db.models import get_app, get_models
+from django.core import serializers
 
 
 # why this way? how about use `__file__`
@@ -2471,6 +2473,15 @@ def build_all():
     build_pokemons()
     build_encounters()
     build_pal_parks()
+
+def dump_all():
+    app = get_app('pokemon_v2')
+    for model in get_models(app):
+        data = serializers.serialize("json", model.objects.all())
+        file_name = os.getcwd() + "/pokemon_v2/fixtures/" + str(model).split("'")[1].split(".")[2] + ".json"
+        out = open(file_name, "w")
+        out.write(data)
+        out.close()
 
 if __name__ == '__main__':
     build_all()


data/v2/build.py
```diff
@@ -2472,5 +2474,14 @@ def build_all():
     build_encounters()
     build_pal_parks()
 
+def dump_all():
+    app = get_app('pokemon_v2')
+    for model in get_models(app):
+        data = serializers.serialize("json", model.objects.all())
+        file_name = os.getcwd() + "/pokemon_v2/fixtures/" + str(model).split("'")[1].split(".")[2] + ".json"
+        out = open(file_name, "w")
```
nimaje, line 2482: you should use the `with` statement:

``` python
with open(…) as out:
    out.write(…)
```

this closes the file even if there is an exception

nimaje:
A problem with this improvment is you normally only build the db if there is new data.
There is normally no need for `make wipe_db`.
This would only help if there is a way to get the fixtures without using the db-buildscript.

@phalt could provide these fixtures as a tarfile for developer which need the data.
But he could also provide a sqlite3 version of the data, this would only need the download time.

data/v2/build.py
mjhayes412, line 2482: Oooo, thank you! I did not know that. I'll take all the tips I can get :)

data/v2/build.py
```diff
@@ -2472,5 +2474,14 @@ def build_all():
     build_encounters()
     build_pal_parks()
 
+def dump_all():
+    app = get_app('pokemon_v2')
+    for model in get_models(app):
+        data = serializers.serialize("json", model.objects.all())
+        file_name = os.getcwd() + "/pokemon_v2/fixtures/" + str(model).split("'")[1].split(".")[2] + ".json"
```
nimaje, line 2481: you should use str.format():

``` python
filename = '{}/pokemon_v2/fixtures/{}.json'.format(os.getcwd(), model.__name__)
```

I didn't test it but I assume you try to parse the `__name__` property

mjhayes412, line 2481: ah geez, :::palm to head:::  that was there the whole time?

Do you think it's wise to separate the models into their own files?  You mentioned @phalt possibly providing a fixture as a tarfile.  I can't tell if it would be easier to have one giant fixture, or smaller chunks of individual fixtures.  Your thoughts?

nimaje:
I think it would be easier to download only one file.
I don't know if it would be better to create only one file.
This should create only one huge fixure file:
```python
with open('fixure.json', 'w') as out:
    out.write(serializers.serialize("json",
        itertools.chain.from_iterable(model.objects.all()
            for model in get_models(get_app('pokemon_v2')))))
```

data/v2/build.py
nimaje, line 2481: Why did I keep `os.getcwd()`? You don't need it. Use a relative filename.

``` python
'pokemon_v2/fixtures/{}.json'.format(model.__name__)
```

phalt:
@mjhayes412 Separate files would make it easier to manage!

@mjhayes412 @nimaje I love this open collaboration, thanks for making me remember why running this project is worth it :)

mjhayes412:
@phalt Thanks for running this project! I had never even heard of Django, Digital Ocean, or ever really use GitHub until I saw the bottom of pokeapi.co/about.  Now I feel like a kid in a candy store (GitHub), reading Django Docs every night, and I just had my first video interview for DO (crossing fingers!)

But yea, this project is awesome.  I think Pokemon is an incredible gateway for students to become interested in software programming.  That's what inspired me.

So what's your end goal?  To have a centralized source of truth for all Pokemon data?  Or perhaps to have one of the foundations for "APIs of APIs" where we can have Pikachu tweet a pizza emoji to Dominos and then post it on your buddy's wall...

...oh man that would be cool.

mjhayes412:
@nimaje 
image

image

phalt:
@mjhayes412 The end goal is to keep this running as a resource for both Pokémon enthusiasts, and also as a tool for teaching people about APIs. It's currently used by many schools in the UK, as well as a lot of coding bootcamps around the world :)

phalt:
@mjhayes412 this pull request needs to be updated with the current master branch.

Can you also update the readme to describe the actions needed to use the new build script?

Issue #154, opened 2016-03-26
MemoryError on building pokemon_v2_pokemonmove


AlecGW:
It's happened the last 3 times I've tried to build this database and I can't manage to fix it. I'm inexperienced with this kind of stuff, so I have no idea what to do with this. The only thing that I've found that causes this otherwise is setting 'DEBUG' to true.
Here's the error log:

```
...
building pokemon_v2_pokemonmove
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "I:\PokeAPI\Newfolder\pokeapi-master\data\v2\build.py", line 2471, in build_all
    build_pokemons()
  File "I:\PokeAPI\Newfolder\pokeapi-master\data\v2\build.py", line 2209, in build_pokemons
    model.save()
  File "C:\Python27\lib\site-packages\django\db\models\base.py", line 591, in save
    force_update=force_update, update_fields=update_fields)
  File "C:\Python27\lib\site-packages\django\db\models\base.py", line 616, in save_base
    with transaction.commit_on_success_unless_managed(using=using, savepoint=False):
  File "C:\Python27\lib\site-packages\django\db\transaction.py", line 300, in __enter__
    connection._start_transaction_under_autocommit()
  File "C:\Python27\lib\site-packages\django\db\backends\sqlite3\base.py", line 466, in _start_transaction_under_autocommit
    self.cursor().execute("BEGIN")
  File "C:\Python27\lib\site-packages\django\db\backends\utils.py", line 88, in execute
    'time': "%.3f" % duration,
MemoryError
>>>
```

I get the feeling it's part of an installation I did badly, but I'm not sure and I've re-installed things to try and getting it to work, but I'm not sure...

mjhayes412:
I'm having the same issue.  Perhaps by serializing the csv files into a json file, this could be resolved using django loaddata.  I'll tinker with it and let you know if it works for me.

OrangeCrush:
@mjhayes412  @AlecGW I'm curious- what are the specs of your build machine??  Also - what platform are you building on? (Windows, Linux, ...etc)  

Just curious as I have built the project hundreds of times (Linux + Docker) without seeing this error.

AlecGW:
Oh yeah I forgot to mention my build in the original post. I'm running Windows (10) so that's probably the problem there. I was considering setting up a Linux VM, but my computer has been acting up lately, so I don't want to try anything like that at the moment.
[...]

PR #169, opened 2016-04-25
Hayes branch

mjhayes412:
Rearranged how build.py works which decreased load time by 50.6% simply by pushing each new object in an array (instead of saving) and afterward using the bulk_create command to save all at once.

Also converted data into json files to take advantage of Django's loaddata functionality which is  76% faster than the original build.py 

json files are saved in pokemon_v2/fixtures per model for easier management.


phalt:
@mjhayes412 This branch is actually too big to accurately review. Could you split it into a few less PRs with are easier to look at? I'd make the json fixtures a separate PR, and the code changes a different one.

nimaje:
@mjhayes412 some files are readed twice (should be 10 after my counting for #154)
And I am against putting autogenerated files under versioncontrol.

`build_generic` could do someting like this:
```python
class2models = collections.defalutdict(list)
for data in daten:
    for model in data_to_model(data):
        class2models[model.__class__].append(model)
for model_class, models in class2models.items():
    model_class.objects.bulk_create(models)
```
this would solve the multiple reading

I don't think you want/should use `'w+'` as open mode (in `dump_one`, `dump_all`)
In `dump_all` you use a `with` statement and `close` (L1655)

mjhayes412:
@nimaje Good points.  I knew about the twice read files for some and I thought about you when I came across them.  I figured I could come back to those 10 cases are refine it but I wanted to keep moving forward. I hope I was going in the direction you were intending.  It made the page look much cleaner.

Ah, I thought I took out that `w+` first.  I ran into a issue where it wouldn't create the file if it didn't exist but I realized I had a `/` at the start of my file name.

Why are you against putting autogenerated files under source control?  It's basically the same thing as providing the csv data files, except the json files load much, much faster.  

@phalt Thanks for the consideration! I'll try and break them up but I may try to improve them with nimaje's suggestions. Also, I should've ran through all the checks before submitting.  I don't have ci/circleci connected and when I ran `make test` I had some errors pop up.

image

phalt:
@mjhayes412 for that error just run `make install` to install the new dependencies.

@nimaje raises a good point, we could auto generate them on the server / CI and then we can just save some space, making the project smaller. Considering we'd need to update it each time the data changes as well...

Makefile
```diff
@@ -1,27 +1,27 @@
 install:
-	pip install -r requirements.txt
+    pip install -r requirements.txt
```
nimaje, line 2: `make` works only with hard tabs
simple demo:

``` sh
$ echo "all:\n    @echo 'test'" > Makefile && make all
Makefile:2: *** missing separator.  Stop.
$ echo "all:\n\t@echo 'test'" > Makefile && make all                                                                                                                              :(
test
```

PR #183, opened 2016-05-05
Added Docker Compose support for production

ThisIsAreku:
This PR add docker-compose support. It allows you to deploy a full production-ready environment.

- It start **gunicorn** with a number of thread depending on the number of cpu.
- Nginx is used as a front proxy.
- Postgres and Redis data are stored in docker data volumes.
- _/assets/_ and _/media/_ directories are not forwarded to python and are processed directly by nginx

I took the liberty to update gunicorn version in the requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,42 @@ Run the container on host port 8000
 docker run -d -p 8000:8000 pokeapi
 ```
 
+
+## Docker Compose
+
+There is also a multi-container setup, managed by [Docker Compose](https://docs.docker.com/compose/). This setup allow you to deploy a production-like environment, with separate containers for each services.
+
+Create data volumes for Redis and Postgres
+```
+docker volume create --name=redis_data
+docker volume create --name=pg_data
+```
+
+Start the process using
+```
+docker-compose up
+```
+You can specify the ```-d``` switch to start in detached mode.   
+This will bind port 80 and 443. Unfortunately, unlike the ```docker``` command, there is no command line arguments to specify ports. If you want to change them, edit the ```docker-compose.yml``` file.
+
+After that, start the migration process
+```
+docker-compose exec app python manage.py migrate
+```
+
+And then, import the data using the shell
+```
+docker-compose exec app python manage.py shell
+```
+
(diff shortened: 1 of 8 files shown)


nimaje:
This nginx config file doesn't look specific for docker.
Why is it in this subdir and why is it named `pokeapi.conf`?
These numbers in `gunicorn.py.ini` seem a little bit random.
And @phalt why is `wsgiref==0.1.2` a dependency  `wsgiref` is since py2.5 in the standardlib and the pypa version seems not to be python3 compatible.

ThisIsAreku:
`pokeapi.conf` is added to the nginx container on docker build. The configuration is adapted to proxy the requests to the app container, I don't understand how it should be more specific to docker.

And the gunicorn configuration is from the docs: http://docs.gunicorn.org/en/latest/configure.html#configuration-file

nimaje:
The question was why it is in this subdir (`docker/`) though not be docker specific.
It could be somewhere else named `nginx.conf` and everyone should see that this is a nginx config file not specific to docker.
After [the docs](http://docs.gunicorn.org/en/latest/settings.html#max-requests) max_requests is to limit the damage of memory leaks (should be not needed if there are no memory leaks) and not set in the config you linked.

ThisIsAreku:
True, I'll move it elsewhere. But as _app_ is hardcoded in the config file, I don't think you can use it easily in a classic nginx installation (ie. you'll have to edit the hosts file to point _app_ to localhost)

phalt:
@nimaje on the `wsgiref` - I've not updated the dependencies in a while. These are the `pip freeze` results from way back when I was building it.

phalt:
@ThisIsAreku looks super good. What's the purpose of this vs the development version that we have in docker?

ThisIsAreku:
This version is more oriented to production system, with a proper isolation between services. Also, (IMO) it's easier to manage an application with _docker-compose_, you don't have to configure manually your server.

phalt:
Okay, I really wanna merge this! 👍 Let's get another person to check it off first

sargunster:
👍

PR #191, opened 2016-05-24
Use empty list instead of null for evolution details

sargunster:
Minor fix related to my earlier PR #190, empty list should be returned instead of null.


phalt:
👍

PR #206, opened 2016-06-04
Allow negative IDs

### LINKED ISSUE #145: Unknown move ailment can't be found by id, only by name
`/move-ailment/-1/` is not found, but `/move-ailment/unknown/` works fine.

sargunster:
* Allow requests with negative IDs to fix `/move-ailment/-1/` (#145)


phalt:
👍

Issue #146, opened 2016-03-15
500 Internal server error on some move categories


sargunster:
Any `move-category` that includes a `+` in its name causes a `500 Internal server error` when called by name. Example: `/move-category/damage+heal/`

sargunster:
It seems any request with a plus in the id/name part will cause the 500 error. Example: this request which should return 404 actually returns 500: http://pokeapi.co/api/v2/pokemon/a+b/. 

Will look into this further.

PR #207, opened 2016-06-04
Fix for dealing with name format

sargunster:
* Allow '+' in names to fix `/move-category/damage+ailment/` (#146)
* Return 404 instead of throwing an exception when it fails


phalt:
👍 

Gonna need a rebase though!

sargunster:
Ah, didn't think about that. I should've bundled these two PRs. Will do the rebase.

PR #208, opened 2016-06-04
Make clean should remove .pyc files in subdirectories too

### LINKED ISSUE #164: make clean
The `make clean` command is supposed to remove compiled python files to force rebuilds and is given the option `-rf` yet it doesn't removed subfiles.

image

sargunster:
fix #164


phalt:
👍  👍

PR #211, opened 2016-06-05
Link moves to machines

### LINKED ISSUE #198: Link moves to machines
Currently there is no way to link TMs/HMs to the moves that they teach. One way to do this would be to provide a nullable field in the items column that is a named reference to the move it teaches.

sargunster:
fix #198 

Show which move is taught by a TM.


sargunster:
Failing test message:

>    django.db.utils.OperationalError: no such table: pokemon_v2_itemtaughtmove

Will investigate.

sargunster:
There, I think it's ready now. You should definitely squash these commits when you merge though to avoid polluting the history.

There's a new resource `/machine/{id}`, and a list field in `/move/{id}` and `/item/{id}` to point to the relevant machines. It also includes updated documentation for the new resource / fields.

pokemon_v2/serializers.py
```diff
@@ -2030,9 +2054,30 @@ class Meta:
             'stat_changes',
             'super_contest_effect',
             'target',
-            'type'
+            'type',
+            'machines'
```
phalt: Python tuples should end each item with a `,`

PR #212, opened 2016-06-05
Redo PokemonEncounterView

### LINKED ISSUE #210: Version details in Pokemon encounters
Seems like the `version_details` field in calls to `/pokemon/{id}/encounters` has gone completely blank recently.

sargunster:
fix #210 

Rewrote this function to fix the bug.


pokemon_v2/api.py
```diff
@@ -413,64 +413,70 @@ def get(self, request, pokemon_id):
...
-
-        # all encounters associated with location area
-        all_encounters = Encounter.objects.filter(
-            pokemon=pokemon).order_by(
-                'location_area').values_list('location_area', flat=True)[:200]
+            raise Http404
+
+        encounter_objects = Encounter.objects.filter(pokemon=pokemon)
+
+        area_ids = encounter_objects \
```
phalt: Let's not use `\` syntax, I think PEP8 prefers an alternative way of formatting. See the guide on line length: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

pokemon_v2/api.py
```diff
@@ -413,64 +413,70 @@ def get(self, request, pokemon_id):
...
+            .order_by('location_area') \
+            .values_list('location_area', flat=True) \
+            .distinct()
```
phalt: I think `distinct` needs to come before `values_list`, otherwise this'll do a comparison after the fact it's made an SQL query, whereas SQL DISTINCT is faster.

PR #214, opened 2016-06-07
Update PokeKotlin link in docs

sargunster:
Forgot to do this when the move was made


phalt:
Nice 👍

phalt:
We should try and source more packages from ones that are maintained by people

sargunster:
Do you know which of the others are currently maintained?

PR #220, opened 2016-07-04
move flavor texts now available

sentwistler:
I noticed that the flavor texts for moves were not currently being provided by the API, but that the data was already present. I added functionality to return them (modeled after the species flavor text as per #111, except that I didn't put in the sections of documentation equivalent to the ones removed in 4bc45830 about species flavor text). The new assertions in `test_move_api` are also modeled after those in #111.

Let me know if you need me to change something! I tried to follow the style used in similar places in the code, but being new to this code and having no prior experience with Django, there may be some areas that can be improved.

### DIFF
--- a/pokemon_v2/serializers.py
+++ b/pokemon_v2/serializers.py
@@ -2011,16 +2011,17 @@ class MoveEffectChangeSerializer(serializers.ModelSerializer):
         model = MoveEffectChange
         fields = ('version_group', 'effect_entries')
 
+
 class MoveFlavorTextSerializer(serializers.ModelSerializer):
 
     flavor_text = serializers.CharField()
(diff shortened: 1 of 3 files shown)


sargunster:
Looks like it failed the lint check. You can check the log to see exactly what changes it wants. Also, the docs should be updated for the new field.

sentwistler:
Ah, okay, wasn't sure about the documentation since the documentation for the species flavor text seems to have been removed

sargunster:
Looks good to me :+1: 

@phalt ?

Also, about the species flavor text, good catch. I must have removed it in that commit by accident. Will fix it soon.

phalt:
Nice 👍

PR #229, opened 2016-07-10
Be consistent in use of parens for list type parameter

relrod:
Makes automatically parsing these tables slightly easier by removing an
edge case

Signed-off-by: Ricky Elrod <ricky@elrod.me>


sargunster:
Thanks for catching the little documentation bugs!

PR #230, opened 2016-07-10
Fix type error for Pokemon/location_area_encounters

relrod:
The API returns a list, not a string

Signed-off-by: Ricky Elrod <ricky@elrod.me>


sargunster:
Actually, it returns a path (string) to a list resource

http://pokeapi.co/api/v2/pokemon/10/

relrod:
@sargunster ah, then the example here is wrong: https://pokeapi.co/docsv2/#pokemon

That is what I was testing with.

sargunster:
Yeah, the example is outdated. Feel free to fix it.

relrod:
@sargunster I'll do future testing with actual API responses. And if I find some time, I'll go through the examples and update them as needed.

PR #240, opened 2016-07-21
Nginx 4cpu configuration

Naramsim:
Since @phalt once said that the Pokeapi machine is an 80$ droplet this means that it has 4 CPUs. [DigitalOcean link](https://www.digitalocean.com/pricing/).

This means that the current Nginx installation is not working at its full potential.

This PR will let Nginx
* use 4 processors 
* handle requests in RAM instead of writing them on HDD
* gzip content before sending it (JSONs, when zipped, are lot lighter)
* cache static content for 3 months (static === Pokeapi website, not API)
* cache JSON responses for 1 minute (Pokemon data does not change very often)

### Note
I need some tester since I am on Windows. I've only tested if Nginx starts correctly, and it does. But I haven't tested it while Pokeapi running.


sargunster:
Nice! I should be able to test it out this weekend.

Resources/nginx/nginx.conf
```diff
@@ -42,6 +58,20 @@ http {
...
+        
+        location ~* \.(jpg|jpeg|gif|png|css|js|ico|xml)$ { # cache all these extensions for 5 days, disable logging
+            access_log        off;
+            log_not_found     off;
+            expires           5d;
+            add_header Cache-Control "public";
+            add_header Pragma public;
+        }
+        
+        location ~* \.(json)$ { # cache JSON for 1 minute, consider to raise it at least to 1 hour
```
ThisIsAreku: `\.(json)$` can be simplified to `\.json$`

ThisIsAreku:
Note that the expires on json will apply only if the URL ends with _.json_ but many wrappers and clients use a path-ending url (like _http://pokeapi.co/api/v2/pokemon/1/_ )

Naramsim:
@ThisIsAreku Whoa, You are completely right. I should search how to manage that case (which is how everyone uses the API... :smile: )

@sargunster bu sure to test it on a Unix machine (epoll only works on Unix) and to have at least 4 cores. Thanks!

sargunster:
Will do on Ubuntu 16.04 w/ i7 6700K .

sargunster:
When caching with Nginx i just put something like this in the `server` block to cache everything:

```
            proxy_cache api;
            proxy_cache_methods GET;
            proxy_cache_valid 200 301 302 404 1d;
            proxy_cache_key "$scheme$request_method$host$request_uri";
            add_header X-Proxy-Cache $upstream_cache_status;
```

phalt:
Incredible stuff @Naramsim - when people approve this (nginx config is new to me!) I can merge and spin up a test server for us to test this on. How does that sound?

ThisIsAreku:
@Naramsim In fact, urls like http://pokeapi.co/api/v2/pokemon/1.json works, and will match the `location` rule 

@sargunster This way of caching by nginx is IMO a good solution to cache response, but I think Cloudflare's cache might be even faster.

Naramsim:
@sargunster should I add your code to the PR?

This one?
```
proxy_cache api;
proxy_cache_methods GET;
proxy_cache_valid 200 301 302 404 1d;
proxy_cache_key "$scheme$request_method$host$request_uri";
add_header X-Proxy-Cache $upstream_cache_status;
```

sargunster:
Something similar to fix the caching issue, yeah. Adapt it for this situation (may not need the header, may not want to cache 404, may not want to change the key, etc). The main thing is to setup caching for anything under `/api`, regardless of extension.

Naramsim:
I've ended up putting the cache control inside the `/api/` location.  Since I don't want to use a proxy server for caching. 
This PR is related only to client-side browser caching.

sargunster:
Just tried it out, I can confirm there are four worker processes and that my browser is caching requests.

Naramsim:
We can boost Nginx by turning off its logging with`access_log off;` in `/api/`
@phalt are you using logs? or can we disable them?

PR #244, opened 2016-07-25
Simple node.js wrapper for Pokéapi

palanik:


Naramsim:
I think people should use [this](https://github.com/PokeAPI/pokedex-promise-v2) wrapper. Since it is complete and auto-caches the resources. It's a big deal here.

If you want you can open issue there and we will try to resolve them

PR #260, opened 2016-08-17
Consistency and URL fix

didii:
Fixed consistency: the table of the object `Machine` did not have a header.

Fixed URL: the table of `Move` mentioned `Move` as the type of `flavor_text_entries` when in reality it is of type `list MoveFlavorText`.

I assume the edited page gets published to the [docs page](https://pokeapi.co/docsv2)?


Naramsim:
Yes, there is a markdown interpreter in docsV2.

PR #261, opened 2016-08-22
badges

Naramsim:
Added 2 badges.

The ghit.me badge will tell us how many users are visiting the repo :smile:


Naramsim:
The sooner we merge, the sooner we will see some stats :smile:

PR #265, opened 2016-09-05
Update berry_firmness_names.csv Spanish names

AlexanderZon:


sargunster:
:+1:

phalt:
These might have to be closed - we do not want to diverge from the main dataset, as this is just a copy of Veekun's

sargunster:
Yeah, this and #264 should be sent over to veekun/pokedex .

PR #266, opened 2016-09-10
Rewording, typos and what not

marcus-crane:
Primarily just some rewording of sentences that didn't flow great.

There's one bug fix to the Stripe payment image, which currently creates a mixed content request error. The imgur url is fetched via http when it should ideally be fetched over https as it's available.


sargunster:
👍

PR #305, opened 2017-09-28
Sun and moon data

Naramsim:
In this PR you will find
* Data from Veekun for Sun and Moon edition
* A quick fix for build_moves()
* Dependencies versions updated
* Links to our official wrappers

I personally tested these changes on my machine using Docker Compose

I would like @tdmalone @sargunv and @phalt to test it.


tdmalone:
Apologies I have not tested this yet @Naramsim. I'm not sure if I'm going to have the time in the near future to jump on this. I will try my best but please don't wait on me or anything!

sargunv:
I'll try to do it today, but if I can't get to it please don't wait on me either!

phalt:
Hey all. I've been away in Japan on holiday for two weeks (visited a Pokémon center and got healed too!) - I'd like to try and look at this, this weekend.

@Naramsim has this been tested against a "built" service - like the production one? If I run the new commands will it overwrite data or will it just append the new content?

Naramsim:
Hi @phalt! 
I didn't test it on a production server environment, I only tested it on my own, using Docker on my Win10 PC. I didn't test it without Docker. So maybe it is a good idea to create a 5€ DigitalOcean droplet and deploy it as you used to (I thought on the prod server was Docker).

I tested it directly with the new data, so I don't know whether the Postgre data is overwritten or appended. But I think that looking at the code will solve this problem. For example, the build_moves() overwrites everything: https://github.com/PokeAPI/pokeapi/blob/8ce2bedf063207ea45a20143d5d7ac87f800cfbd/data/v2/build.py#L964 and so I guess it is for all the other tables.

I'm not sure if the statistics will be overwritten, those are not built under `build_all()`, so I think only Pokemon data will be overwritten entirely.

Maybe on the official server we could `git pull`, and then only `build_all()` in a shell. But before do a snapshot of the machine (I think they are free). Even without restarting the server with the new dependencies, just only re-building the data.

But of course, I think that re-build everything should be the right decision, lots of dependencies had upgraded and should give us better performance and security

phalt:
You read my mind! I'll definitely `git pull` and then run `build_all()` in a shell :)

I do remember doing it in the past with no issues. I'll double check the code when I go and run it to see if we're overwriting or not.

Issue #312, opened 2017-10-28
Documentation shows HTTP instead of HTTPS


JordanMajd:
The home page docs show `http://pokeapi.co/api/v2/` as the URL but hitting it returns a `301 Moved Permanently` to `https://pokeapi.co/api/v2/`. 

Normally this is not much of an issue as most tools automatically handle the redirect, but some do not.  `curl` is one of these.  

I normally use `curl` to test out an API and was curious when `curl http://pokeapi.co/api/v2/` did not print anything. I ran `curl -v http://pokeapi.co/api/v2/` which returned:

```
* Hostname was NOT found in DNS cache
*   Trying 2400:cb00:2048:1::6812:3b8f...
* Connected to pokeapi.co (2400:cb00:2048:1::6812:3b8f) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.35.0
> Host: pokeapi.co
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Date: Sat, 28 Oct 2017 02:34:03 GMT
< Transfer-Encoding: chunked
< Connection: keep-alive
< Cache-Control: max-age=3600
< Expires: Sat, 28 Oct 2017 03:34:03 GMT
< Location: https://pokeapi.co/
* Server cloudflare-nginx is not blacklisted
< Server: cloudflare-nginx
< CF-RAY: 3b4a7e6b756d5011-DEN
<
* Connection #0 to host pokeapi.co left intact
```

Displaying `https` instead of `http` in the docs would allow the client to directly access the resource without the client needing to redirect the `301` response. 

I'd be happy to submit a PR if this is a desired behavior.

tdmalone:
I _think_ the https was a recent move, so I’m sure updating the docs wouldn’t go astray!

As a side note - and you may know this anyway - you can pass the ‘--location’ switch to curl to cause it to follow redirects.

phalt:
It was very recent. A PR will a grep/replace of http -> https would be easily accepted :)

Naramsim:
Docs now always use HTTPS

PR #314, opened 2017-11-30
Update http -> https for pokeapi endpoints in home.html

HipsterBrown:
Fixes https://github.com/PokeAPI/pokeapi/issues/312

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ A RESTful API for Pokémon
 
 LICENSE: [BSD](https://github.com/PokeAPI/pokeapi/blob/master/LICENSE.rst)
 
-http://pokeapi.co
+https://pokeapi.co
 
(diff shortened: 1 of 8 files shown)


Naramsim:
Hi @HipsterBrown, thanks for the PR.

We can merge it, but before merging I would suggest a thing. Maybe in future, we will drop Cloudflare and thus run only over HTTP protocol. So it would be nice that if pokeapi.co is loaded using HTTPS it shows https:// and if it is loaded using HTTP is shows http://

We can use `location.protocol` to determine which protocol is the page loaded

HipsterBrown:
@Naramsim That's an interesting case to think about. 

Using something like `location.protocol` would only catch the cases on the site pages, so some manual work would still be needed to update the markdown pages across this repo. 

I'm curious about how the `url` resources are generated for the API because that seems to stay updated. I.e the `results` from `/api/v2/pokemon/`:

```json
{
	"count": 949,
	"previous": null,
	"results": [
		{
			"url": "https://pokeapi.co/api/v2/pokemon/1/",
			"name": "bulbasaur"
		},
		...
	],
	"next": "https://pokeapi.co/api/v2/pokemon/?limit=20&offset=20"
}
```

How does that `url` key stay updated?

Naramsim:
Based on the client request. When you could query the API using HTTP the returned URLs where http:// if you were using http, https:// if you were using https. Now it returns only https because is the only protocol allowed.

Naramsim:
Closing since the frontend moved here https://github.com/PokeAPI/pokeapi.co

Issue #332, opened 2018-04-17
encounters return is different than other endpoints'


jrubinator:
Right now the encounters endpoint lives at `https://pokeapi.co/api/v2/pokemon/{id}/encounters`

For a given id, the URL returns a list, rather than a hash. On the other hand, the endpoint URLs (eg. `/api/v2/pokemon`) return very similar data, but prefix that data with `results:` Would it make sense to do this here? Unfortunately, that's not a backwards-compatible change.

I suggest this because I believe it would make it easier for API clients to handle the response - see eg.
https://github.com/GregHilmes/pokebase/issues/10

Naramsim:
Hello, back in time the encounter endpoint wasn't there, but part of `/api/v2/pokemon/{id}/`. It was created to lighten the responses of this pokemon endpoint. So it was a quick improvement. Right now the repository is in a stalemate state, and probably no one will implement this alignment with `/api/v2/pokemon/`

jrubinator:
Interesting - that migration makes sense.

I guess I was wondering if the following change makes sense at the last line of `pokemon_v2/urls.py`:
```
-        return Response(encounters_list)
+        return Response({'results': encounters_list})
```

I haven't worked with Django before, or I might be able to suggest something more mature (with proper pagination). That sounds like the alignment you doubt will happen though.

Let me know what you think.

Naramsim:
Yeah, probably is that simple fix 😄 So maybe you can open a PR and put me in the 'request review' and one day maybe it will be merged, but this change obviously could break lots of applications. So we have to be careful about deciding what to do.

Naramsim:
Closing this issue and marking it for a future v3 version.

Right now we cannot change the endpoint format since it's used by many people. We acknowledge that your proposed solution would be the best option.

PR #335, opened 2018-04-21
dictionary from encounters subresource (#332)

jrubinator:
This makes it consistent with the every other resource
Unfortunately this isn't tested or testable in the current framework,
as distinct() is not supported by the sqlite3 backend

### DIFF
--- a/pokemon_v2/api.py
+++ b/pokemon_v2/api.py
@@ -487,4 +487,4 @@ class PokemonEncounterView(APIView):
                 'version_details': version_details_list
             })
 
-        return Response(encounters_list)
+        return Response({'results': encounters_list})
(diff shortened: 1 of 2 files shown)


AgustinRamiroDiaz (approved):

TheSatanicFlame (approved):

omniuni:
The encounters resource, as it is, is really weird. I'd love to see this PR merged! One note about this; because it also changes the endpoint, this should be able to be updated so that it simply _adds_ a new endpoint that operates as expected but leaves the old endpoint in the old format for compatibility.

phalt:
Closing as this is old and I want to tidy things up.

PR #340, opened 2018-06-01
New PHP Wrapper, with lazy-loading and POPOs

lmerotta:


Naramsim:
Hi, 
thanks for this wrapper! It looks awesome, I also suggest you to modify our readme to include it. Meanwhile you could add some tests to your module :wink:, that would be awesome.

lmerotta:
Hi,
Thanks for the feedback ! Tests are on their way 😄 Should I add the wrapper in the official wrappers section ?

Naramsim:
Maybe you can add it when you finish the missing endpoints (Contests, Pokéathlons). Is there a specific reason for why you left them out?

lmerotta:
They are under development actually, I released the wrapper because it was complete enough to be published. The missing endpoints will be added in the 1.1.0 release in the following days.

Maybe we can leave this pull request open, and I’ll update it once the wrapper hits 1.1.0, are you okay with this ?

Naramsim:
I think we could wait a second for @tdmalone, I really trust him on PHP development.

I really like the structure of your wrapper and the _proxy_ function.

PR #351, opened 2018-09-06
Improve performance of database build script (WIP)

cmmartti:
Along the same lines as #169, this PR improves the performance of the v2 database build script by taking advantage of Django's `bulk_create` function. Initial tests indicate build times to be a small fraction of before.

The modified build script calls bulk_create in batches of 200 records, and also sets foreign keys directly (e.g. `language_id = int(info[1])`) rather than getting the object with that key and assigning it to the model (e.g. `language = Language.objects.get(pk = int(info[1]))`).

Setting foreign keys directly has the downside that it will fail if that foreign key doesn't already exist, meaning it won't be possible to build just a part of the DB like we can now (so you will only be able to call build_all()), but it's all around a much faster process now so that shouldn't really matter.

This is a work in progress. Only half of the build script has been converted so far, but it is still functional.

### DIFF
--- a/data/v2/build.py
+++ b/data/v2/build.py
@@ -76,18 +76,26 @@ def clear_table(model):
             "SELECT setval(pg_get_serial_sequence(" + "'" + table_name + "'" + ",'id'), 1, false);")
 
 
-def process_csv(file_name, data_to_models):
-    daten = load_data(file_name)
-    next(daten, None)  # skip header
-    for data in daten:
-        for model in data_to_models(data):
-            model.save()
-
-
 def build_generic(model_classes, file_name, data_to_models):
+    models = {}
     for model_class in model_classes:
         clear_table(model_class)
-    process_csv(file_name, data_to_models)
+        models[model_class] = []
+
+    daten = load_data(file_name)
+    next(daten, None)  # skip header
+
+    for data in daten:
+        for model in data_to_models(data):
+            models[type(model)].append(model)
+
+            # Limit the batch size
+            if len(models[type(model)]) > 200:
+                type(model).objects.bulk_create(models[type(model)])
+                models[type(model)] = []
+
+    for model_class, models_list in models.iteritems():
+        model_class.objects.bulk_create(models_list)
 
 
 def scrubStr(str):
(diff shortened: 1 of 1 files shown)


sargunv:
This looks great; I'd be happy to review once it's fully converted!

As an aside, do you think it's possible to automatically generate the highly repetitive parts of the build script using the model definitions in pokemon_v2/models.py? That's probably more suited to a separate PR, but it would make modifying the build process in the future much easier.

cmmartti:
@sargunv
I thought about it, but the csv column names don't always match up to the model attribute names, and you would still need a way to figure out which file goes to which model. Also, some models (like image sprites) require custom processing.

In any case, all of the procedural code has been pulled out into helper codes, so the majority of the build script is a simple mapping between models and the corresponding records from a specific file.

cmmartti:
@sargunv
I have completed the modification of the build script, and successfully ran it against a PostgresQL database and an SQLite database (both on a Surface Pro 3 running Windows 10). Building against the PostgresQL db took 5 minutes 40 seconds on my machine, while SQLite took 26 minutes 18 seconds. Both are huge improvements.

I have never been able to build against an SQLite database without the process failing early on in the script, and building against a PostgresQL database was always an overnight job. I don't know exactly how long it took, but something like 5 or 6 hours. If we assume 5 hours, then that's an improvement of 6000%.

I have also updated the README.

sargunv:
Sounds awesome, I'll review over the weekend. 60x improvement is incredible, thanks for doing this!

README.md
```diff
@@ -117,41 +117,9 @@ run the build script with
...
-- build_genders()
-- build_experiences()
-- build_machines()
-- build_evolutions()
-- build_pokedexes()
-- build_locations()
-- build_pokemons()
-- build_encounters()
-- build_pal_parks()
+Each time the build script is run, it will iterate over each table in the database, wipe it, and rewrite each row using the data found in data/v2/csv. Building against a PostgresQL database should take 5–8 minutes, and building against a SQLite database should take about 20–30 minutes.
```
sargunv: I'd recommend adding some machine context to these estimated build times (tested on an SSD or HDD, on which CPU, etc).

data/v2/build.py
```diff
@@ -76,18 +74,26 @@ def clear_table(model):
...
-def process_csv(file_name, data_to_models):
-    daten = load_data(file_name)
-    next(daten, None)  # skip header
-    for data in daten:
-        for model in data_to_models(data):
-            model.save()
-
-
 def build_generic(model_classes, file_name, data_to_models):
+    models = {}
```
sargunv: I feel the term `model` is being overloaded in this function. We have:

 - `model_class` is a class that defines the shape of data in an endpoint
 - `models` contains the batches of items to insert into the database
 - `model` is a row of the csv mapped to an object

Some stuff here should be renamed to make it easier to understand. My recommendations: 

 - `model_classes` stays the same
 - `models` becomes either `batches` or `object_batches`
 - `data` becomes either `raw_data` or `csv_data`
 - `record` becomes either `raw_record` or `csv_row` or `csv_record`
 - `model` becomes either `obj` or `object_record`
 - `models_list` becomes either `batch` or `object_batch`

And perhaps in the whole file: 

 - `data_to_models` becomes either `raw_data_to_objects` or `csv_data_to_objects`

This would follow the standard Django terminology where a "model" refers to the class defining the shape of a table and an "object" is a single record in that table, while also making it more clear at a glance where we have raw data from the csv and where we have objects for the Django models.

sargunv:
Built in about 80 seconds on my machine -- wow. It used to take hours.

cmmartti:
I made the changes you suggested and re-ran the script against PostgresQL and SQLite. For some reason, I got wildly different results with the SQLite tests, twice, probably because I was downloading something in the background the first time which may have slowing slowed down disk access or something. I put the better results in the readme (7.5 minutes each time I ran it), along with my machine's stats.

data/v2/build.py
```diff
@@ -74,26 +74,27 @@ def clear_table(model):
...
+    for csv_record in csv_data:
+        for obj in csv_record_to_objects(csv_record):
+            model_class = type(obj)
+            batches[model_class].append(obj)
 
             # Limit the batch size
-            if len(models[type(model)]) > 200:
-                type(model).objects.bulk_create(models[type(model)])
-                models[type(model)] = []
+            if len(batches[model_class]) > 200:
```
sargunv: Might be good to define this number in a constant at the top of the file

sargunv:
Approved, just need one more person's approval to merge

data/v2/build.py
cmmartti: Good point. In this case, I think it's fine for now, but I'll keep it in mind for next time. 200 isn't a hard number, just a number I decided on based on a couple of quick tests and some googling.

PR #360, opened 2018-09-08
chore - migrate to CircleCI 2.0; resolves #359.

### LINKED ISSUE #359: Migrate to CircleCI 2.0
CircleCI 1.0 [has been deprecated](https://circleci.com/blog/sunsetting-1-0/). We should migrate to CircleCI 2.0.

neverendingqs:


.circleci/config.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  test:
+    docker:
+      - image: circleci/python:2.7
+    steps:
+      - checkout
+      - run:
+          name: Install deps
+          command: sudo make install && sudo make dev-install
```
neverendingqs, line 11: The CircleCI build user doesn't have permissions to install system-level deps (see https://circleci.com/gh/PokeAPI/pokeapi/255 for an example), so I went with the simpliest solution (`sudo`).

sargunv: It's fine for now in CI, but it's probably a good idea to move user facing instructions to include virtualenv. Pipenv or Poetry are two simple options that'll give the user a virtualenv by default. 

That's probably best suited to a separate PR, which can then also update this config to keep CI as similar to the user's environment as possible.

neverendingqs:
@tdmalone - once you're happy with this PR, could you remove the `ci/circleci` status check requirement, and add in the `ci/circleci: test` one instead?

Alongside the requirement to keep branches up-to-date, all other PRs will automatically become unmergeable, which is good in terms of migrating those branches over to CircleCI 2.0.

tdmalone:
Thanks for doing this @neverendingqs!

I might be missing something - but compared to our existing CircleCI config, we seem to be missing the lint step. Is that something we can re-add, or was there a specific reason for dropping it?

neverendingqs:
No I just missed it :(. I won't be able to get to it today - anyone feel free to take over the PR if you want.

Issue #362, opened 2018-09-09
Merging and linting strategies


tdmalone:
Our two main repositories are now pokeapi and [ditto](https://github.com/PokeAPI/ditto/).

To reduce confusion for maintainers, it would probably be a good idea to standardise management of these repos, so I was looking at what is currently set up differently between the two. One item is the allowed merge strategies:

• pokeapi currently allows rebase merging only
• ditto currently allows squash merging only

The remaining option is merge commits.

Does anyone have any concerns or thoughts one way or the other about how we should manage merging? I personally like merge commits, dislike squash merging, and am indifferent about rebase merging - but in the end I don't really mind what we go with; I just love consistency 😆 

Does anyone have any strongly held opinions on this, or otherwise any objections to changing ditto to the same merge strategy as pokeapi, now that it is a part of the PokeAPI organisation? Ping @sargunv as the primary maintainer of ditto. 😄 

(Worth noting it's also possible to allow multiple merge strategies - although I think it might be better to stick to one)

sargunv:
I generally dislike merge commits because the diff is duplicated across two commits (the real one and the merge one) in the commit history, and it pollutes the history a bit with extra commits.

I'm happy with either rebase or squash commits, but I prefer squash so incremental changes in a PR (say, after a review) don't pollute the history.

Should we standardize linting across pokeapi and ditto too? My personal preference here is [black](https://black.readthedocs.io/en/stable/) and [flake8](http://flake8.pycqa.org/en/latest/) configured to ignore PEP8 formatting (W and E error classes).

I've also been experimenting with [Stickler-CI](https://stickler-ci.com/teams) to enforce black and flake8 on PRs in ditto. Overall I like the concept, since it'll automatically apply black and update the PR instead of commenting, but it's also been a bit buggy (example: https://github.com/PokeAPI/ditto/pull/17). I may just use a more traditional CI solution to enforce lints (something I plan on doing anyway to enforce tests).

tdmalone:
I've edited the scope of this issue to cover linting as well - definitely would be good to standardise that as well. I haven't used Stickler before but I love the idea of it - if it works that is, haha.

We currently use flake8 for the main PokeAPI repo; if we added black, would this involve significant reformatting to comply?

sargunv:
Blacks formatting is mostly similar, the main difference is 10% longer line length and lack of configuration options (which I see as a positive). I can open a PR with black applied to get an idea of the change.

sargunv:
[...]

PR #363, opened 2018-09-09
Apply black formatting to pokemon_v2

sargunv:
Per the discussion in https://github.com/PokeAPI/pokeapi/issues/362, here's what black would do to PokeAPI.


sargunv:
Closing this since it's not meant to be merged. It can serve its purpose as an example without being open.

PR #382, opened 2018-10-14
fix: Remove old website and v1 API

cmmartti:
Since the pokeapi.co website has been rebuilt and moved to a separate repo, the website in this repo is obsolete.

This removes all files and code related to the website, including the `alerts` and `hits` apps (which managed website alerts and the hit counter) and all site assets (CSS, JS, and images). All non-API URLs have been removed from the `urls.py` file, and the `settings.py` cleaned up a bit. I'm sure I missed something though, so it would be a good idea if someone else took a look at it too.


Naramsim:
I'm not sure why the location tests are failing.

cmmartti:
Nor I. It works on my machine (of course 😑). However, when I switch from SQLite to Postgres, a different set of tests fails. I'll have to look into this before this can be safely merged.

Issue #301, opened 2017-07-25
Update to Python 3


cweeks12:
I'm working on updating the api to Python 3. I'm not sure how long this will take; it seems simple right now, but I'm just starting on it. I'm opening this issue to let people know I'm working on it. I forked the repo and you can see the progress here: https://github.com/cweeks12/pokeapi.

Naramsim:
Perfect, thanks for your effort.

The main impact on performance will come when upgrading Django dependency. Remember to target Python 3.6

Naramsim:
Also, @cweeks12, if you want you can drop Django, which is a little bit slow, and use [Sanic](https://github.com/channelcat/sanic)

Obviously, you can replace only the API server, not the entire docs server(which uses Django templating)

cweeks12:
I really like what I'm seeing with that Sanic. I think I'll move it all over to that.

I haven't looked at the code a ton yet, but should I move the API v1 to Python 3/Sanic as well? Or should I just focus on moving v2 and dropping v1?

Naramsim:
Nono, v1 is not maintained anymore, just leave as it is now, or remove it, to make things clear

cweeks12:
Okay, I think I've decided that I'm going to use Sanic instead of django for the framework, and then instead of using the built-in django database api, I think I may move over to peewee and just keep the sqlite database intact. Thoughts?

cweeks12:
This [stackoverflow answer](https://stackoverflow.com/questions/39803746/peewee-and-peewee-async-why-is-async-slower) has me leaning towards just regular peewee, instead of the async flavor, but I just want know what everyone else thinks, if I'm on the right track or completely off the rails

Naramsim:
Yep, the important is to have the API working, once it is working we can think about non-API relative things, such as Docs and Landing page. But it will be pretty simple to plug them in since those are HTML pages.

Naramsim:
@cweeks12 keep in mind that on Pokeapi server we are using Postgres, SQLite is just a quick alternative.
https://github.com/PokeAPI/pokeapi/blob/8ce2bedf063207ea45a20143d5d7ac87f800cfbd/data/v2/build.py#L73

phalt:
Few things:

1) Please don't remove V1. It is still the most used part of the API and will break a bunch of old integrations if we remove it. (the price of managing APIs is always maintaining versions).
2) Moving away from Django will be tricky, considering the server configuration is built around it!
[...]

PR #385, opened 2018-10-19
Upgrade to Python 3.7 and Django 2.1.2

cmmartti:
This pull request resolves issue #301.

Simultaneously upgrade Python to 3.7 and Django to 2.1.2. All other dependencies are also updated to their most recent versions. With the help of [2to3](https://docs.python.org/2/library/2to3.html), manually correct all incompatible code. Surprisingly, there wasn't much. Whoever wrote the initial code (@zaneadix and @phalt) did a good job of staying compatible with Python3.

All tests pass. The database build script was also tested and corrected.


cmmartti:
I'm running into a couple of issues with the Flake8 linter. Flake8 hasn't been updated in nearly a year, so it doesn't support the most recent version of pycodestyle, which fixed a bug that throws an error in Python 3.7 (issue at PyCQA/pycodestyle#728).

Also, for some reason, Flake8 isn't respecting the `#NOQA` comment that skips linting on specific lines, and there are a number of modules that use `*` star imports. This could be "fixed" by adding error code `F405` to the ignore list, but doesn't fix the underlying problem, which has me stumped.

Options:
- Disable linting until it works with Python 3.7
- Switch to a different linter
- Downgrade to Python 3.6

Thoughts?

Naramsim:
I'd say to just switch the linter. PyLint should work well. I wouldn't
downgrade Python only because the linter doesn't work.

cmmartti:
Okay, it should be good now. I switched the linter to PyLint and adjusted the default configuration to something sane, then fixed the errors and warnings that were detected.

sargunv:
The close was accidental, looks good. The new linter does make me wonder how close we are to black's formatting now.

PR #395, opened 2018-11-30
Update ISSUE_TEMPLATE with note about data source

cmmartti:
We often get issues that our data is not up-to-date, which is sometimes beyond our control. Hopefully this will direct these requests to veekun.


cmmartti:
It looks like the linter is being a bit overzealous...

PR #406, opened 2019-01-07
fix: Revert docker-compose instructions in readme

cmmartti:


README.md
```diff
@@ -118,16 +118,16 @@ This will bind port 80 and 443. Unfortunately, unlike the `docker` command, ther
 After that, start the migration process
 
 ```sh
-python manage.py migrate --settings=config.docker-compose
+docker-compose exec app python manage.py migrate
```
Naramsim: `--settings=config.docker-compose` is mandatory, otherwise sqlite will be used.

cmmartti: It won't use SQLite, but it will use the wrong PostgresQL settings, which only differ by DB user (root vs. ash). That's why it still worked when I tried it.

Naramsim: Ok, but since we have a https://github.com/PokeAPI/pokeapi/blob/master/config/docker-compose.py it is better to use it :smile:

PR #415, opened 2019-02-11
data in some languages(Korean, Japanese, French, German) added a little bit.

ymin1103:
I made a small data contribution in some languages : Korean(especially), Japanese, French, German.

- version_names : Korean, French, German, Japanese

- generation_names : Korean

- stat_names : Korean

- egg_groups : Korean

- location_names : Korean


cmmartti:
We take our CSV data directly from [Veekun](https://github.com/Veekun/pokedex) without modifying it, so please direct any translations, etc. to them. If they're merged there, those changes will make their way into PokéAPI. Do you know Korean, French, German, and Japanese yourself? If not, where are the translations from?

ymin1103:
Ok, sorry for my ignorance. So do I have to make contribution at Veekun when it comes to translation, etc? 
I know Korean, English, Japanese. French and German translated data were from some wikipedias about pokemon  (ie. bulbapedia)

cmmartti:
Yes, exactly. They'll want to know that the translations are accurate, which is why I asked.

ymin1103:
Thank you for your kindness. 😀

PR #446, opened 2019-08-22
Typo for pikachu(#25)

Deondrede:
New to GitHub, hopefully I'm making correct use of this. Typo in the flavor text description for pikachu


magical:
This is not a typo. You might not be able to see it, but there is a form feed (^L, ascii 12) between "their" and "electricity", representing the page break between the two pages of pokedex text in the early games.

More info: https://github.com/veekun/pokedex/issues/218#issuecomment-339841781

PR #464, opened 2019-12-17
Disables throttling

Naramsim:
Since the entire API is now static files, we don't need anymore any throttling. This simplifies also updating data in the `api-data` repo.

### DIFF
--- a/Resources/nginx/nginx.conf
+++ b/Resources/nginx/nginx.conf
@@ -38,8 +38,6 @@ http {
         server app:8000 fail_timeout=0;
     }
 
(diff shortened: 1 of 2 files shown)


Kronopt:
How is this going to affect the Fair Use Policy described in the pokeapi.co documentation?

Naramsim:
It won't since we are not using anymore this python server to serve the APIs, we are using static hosting on Firebase.

PS: And now with firebase, there is no throttling at all.

Naramsim:
@phalt, can you unlock?

PR #475, opened 2020-02-17
Add Rust wrapper to wrappers list

lunik1:
Hello, I have written a PokéAPI wrapper in Rust. This PR adds it to the list of official wrappers.

All endpoints are supported and API responses are cached.


phalt:
Thanks for the contribution!

PR #485, opened 2020-04-13
Bugfix #163: serialise evolution chains recursively to fix Wurmple evolution

### LINKED ISSUE #163: Wurmple evolution incorrect.
http://pokeapi.co/api/v2/evolution-chain/135/

The Wurmple evolution isn't correct. The "Cascoon" to "Dustox" are out of place. As it is in the current api, the Cascoon object is on the same level as Beautifly. The (simplified) correct evolution is below.

Wurmple evolves_to: [
     Silcoon evolves_to: Beautifly,
     Cascoon evolves_to: Dustox
]

phrasmotica:
This fixes #163 so that Wurmple's evolution is now correctly displayed. I see this is a long-standing issue so hopefully this fix can be merged in! 😄

### DIFF
--- a/pokemon_v2/serializers.py
+++ b/pokemon_v2/serializers.py
@@ -3134,44 +3134,78 @@ class EvolutionChainDetailSerializer(serializers.ModelSerializer):
             pokemon_objects, many=True, context=self.context
         ).data
 
-        chain = entry = OrderedDict()
-        current_evolutions = None
-        evolution_data = None
-        previous_entry = None
-        previous_species = None
+        # convert evolution data list to tree
+        evolution_tree = self.build_evolution_tree(ref_data)
 
-        for index, species in enumerate(ref_data):
-
-            # If evolves from something
-            if species["evolves_from_species"]:
-
-                # In case this pokemon is one of multiple evolutions a pokemon can make
-                if previous_species:
-                    if previous_species["id"] == species["evolves_from_species"]:
-                        current_evolutions = previous_entry["evolves_to"]
-
-                entry = OrderedDict()
-
-                evolution_object = PokemonEvolution.objects.filter(
-                    evolved_species=species["id"]
-                )
-
-                evolution_data = PokemonEvolutionSerializer(
-                    evolution_object, many=True, context=self.context
-                ).data
-
-                current_evolutions.append(entry)
-
-            entry["is_baby"] = species["is_baby"]
-            entry["species"] = summary_data[index]
-            entry["evolution_details"] = evolution_data or []
-            entry["evolves_to"] = []
-
-            # Keep track of previous entries for complex chaining
-            previous_entry = entry
-            previous_species = species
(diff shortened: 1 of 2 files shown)


Naramsim:
Thanks a lot for the contribution!! @zaneadix is the best contributor that can validate this PR, I guess and if he wants. 

@phrasmotica would you mind also to try and write a test that tests your addition? 

https://github.com/PokeAPI/pokeapi/blob/243c6acd2a519108cdb73999e57d9786e14ce725/pokemon_v2/tests.py#L5100

phrasmotica:
@Naramsim sure I can add a test 🙂

Naramsim:
I would like a review from @phalt especially looking at the logic behind the code. I reviewed by myself the PR but I'm no expert in Django.

pokemon_v2/serializers.py
```diff
@@ -3134,44 +3134,78 @@ def build_chain(self, obj):
...
-        current_evolutions = None
-        evolution_data = None
-        previous_entry = None
-        previous_species = None
+        # convert evolution data list to tree
+        evolution_tree = self.build_evolution_tree(ref_data)
 
-        for index, species in enumerate(ref_data):
+        # serialize chain recursively from tree
+        base_chain_link = evolution_tree
```
phalt: Why are you reassigning the var? Makes no sense

pokemon_v2/serializers.py
```diff
@@ -3134,44 +3134,78 @@ def build_chain(self, obj):
...
-                if previous_species:
-                    if previous_species["id"] == species["evolves_from_species"]:
-                        current_evolutions = previous_entry["evolves_to"]
+    # converts a list of Pokemon species evolution data into a tree representing the evolution chain
+    def build_evolution_tree(self, species_evolution_data):
+        evolution_tree = None
 
-                entry = OrderedDict()
+        for species in species_evolution_data:
+            if not evolution_tree:
```
phalt: The code in this IF conditional isn't need and could just go at line 3148 as the default.

pokemon_v2/serializers.py
phrasmotica: Not a clue! Will remove it.

Naramsim:
@phalt, yes, we need to run the updater bot locally on one of our PCs. But I was thinking of creating a technical user (machine user https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users). Then we could move the updater bot to this repository and use the technical user during the deploy phase:

1. PR is merged here on Pokeapi
2. CircleCI runs the updater bot using the technical user and submittes a new PR towards the api-data repo with the new generate data
3. If everything seems right we manually merge the PR in the api-data repo and this will trigger an update on Firebase with the new data.

Sound good?

So we don't actually need to run anymore the updater-bot on our PCs.

phalt:
I am all for moving that task into a CircleCI job yep. Anything like that should be automated away :)

PR #490, opened 2020-05-03
Prevent server error on large integer lookup

### LINKED ISSUE #487: Error when using large integers and proposed fix
Hi PokeAPI team!

Thanks for your awesome API.

My company is developing a service for automated testing of GitHub apps, and while testing it out on some public repos we found the following bug [in the `get_object` function from `api.py`](https://github.com/PokeAPI/pokeapi/blob/3c1115cf0c1bfab21c3900f26b9535e558d510bf/pokemon_v2/api.py#L41). Because sqlite does not support large integers, using a too-large integer as an `id` (ie `/api/v2/version-group/9223372036854775808/`) will cause the server to `500`.

There are a few possible fixes for this on the db-level (ie [this stack-overflow one](https://stackoverflow.com/questions/12150973/django-id-integer-limit)) and an app-level one would be to validate any integers in the query before sending in `get_object` and returning a sensible error code if the integer is too big or small.

We really enjoyed testing PokeAPI with Meeshkan, and if you find this bug report useful, I'd encourage you to sign up for free alpha access at https://meeshkan.com. Thanks again for your great work!

aitorres:
This PR implements an app-solution to issue #487, in which:

> Because sqlite does not support large integers, using a too-large integer as an id (ie /api/v2/version-group/9223372036854775808/) will cause the server to 500.

According to the [IntegerField entry on Django docs](https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.IntegerField) (which is the field type used for primary keys on models), 

> Values from -2147483648 to 2147483647 are safe in all databases supported by Django.

I think it is safe to assume that, for a long time, no data category on the PokéApi will reach more than 2147483647 entries.

The implemented fix simply checks that numeric lookup values are less than 2147483647 in absolute value. If the validation fails, a 404 error is directly raised. Otherwise, the database query proceeds as expected.

A new test has been added to the tests files, in order to check this case. All tests passed successfully in my local environment.

This is my first PR here, hope everything's okay, and of course I'm open to changes and/or suggestions!

If merged, this PR fixes and closes #487

### DIFF
--- a/pokemon_v2/api.py
+++ b/pokemon_v2/api.py
@@ -44,6 +44,10 @@ class NameOrIdRetrieval:
         lookup = self.kwargs["pk"]
 
         if self.idPattern.match(lookup):
+            lookup_id = int(lookup)
+            if abs(lookup_id) > 2147483647:
+                raise Http404
+
             resp = get_object_or_404(queryset, pk=lookup)
 
         elif self.namePattern.match(lookup):
--- a/pokemon_v2/tests.py
+++ b/pokemon_v2/tests.py
@@ -5438,3 +5438,10 @@ class APITests(APIData, APITestCase):
             response.data["pokemon_encounters"][0]["pokemon_species"]["url"],
             "{}{}/pokemon-species/{}/".format(TEST_HOST, API_V2, pokemon_species.pk),
         )
+
+    # ID Range Tests
+    def test_id_range_api(self):
+
+        response = self.client.get("{}/pokemon/{}/".format(API_V2, 2147483648))
+
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)


pokeapi-machine-user:
A [PokeAPI/api-data](https://github.com/PokeAPI/api-data) refresh has started. In 45 minutes the staging branch of [PokeAPI/api-data](https://github.com/PokeAPI/api-data/tree/staging) will be pushed with the new generated data. <br><br> The staging branch will be deployed in our [staging environment](https://staging.pokeapi.co/api/v2/) and you will be able to review the entire API. <br><br> A Pull Request ([master](https://github.com/PokeAPI/api-data/tree/master)<-[staging](https://github.com/PokeAPI/api-data/tree/staging)) will be also created at [PokeAPI/api-data](https://github.com/PokeAPI/api-data/pulls) and assigned to the PokeAPI Core team to be reviewed. If approved and merged new data will soon be available worldwide at [pokeapi.co](https://pokeapi.co/api/v2/).

pokeapi-machine-user:
The updater script has finished its job and has now opened a Pull Request towards [PokeAPI/api-data](https://github.com/PokeAPI/api-data/pulls) with the updated data. <br><br> You can see the Pull Request deployed at our [staging environment](https://staging.pokeapi.co/api/v2/) when [CircleCI deploy](https://app.circleci.com/pipelines/github/PokeAPI/deploy) will be finished (_check the started time of the last build_).

PR #493, opened 2020-05-13
Retrive PR number when not set in the env

Naramsim:
[Apparently](https://discuss.circleci.com/t/circle-pull-requests-environment-variable-not-being-set/29883) CircleCI doesn't inject `CIRCLE_PULL_REQUEST` env variable when a PR is merged into the `master` branch, so we have to retrieve the number of the merged pull request by looking at the last commit message.


Naramsim:
You're right, but the commit last commit message will be always something like this when a PR will be merged:

```
Merge pull request #488 from PokeAPI/circleci-hook
```

You can look at the last commit message here: https://github.com/PokeAPI/pokeapi/commit/d5b43630fabb61b4c9699eff313592b71edbccba

So the pattern is:

```
Merge pull request #{number} from {repo}/{branch}
```

We are safe if the branch does not contain any numbers. But I admit it, we should handle also that case. I'll do it tomorrow.

You have also to know that we cannot directly push on the `master` branch, so the last commit will always be a _merge_ commit.

sargunv:
This seems extremely fragile to me. Someone in the future might unlock the master branch, not realizing it’ll break this workflow. GitHub might change the merge commit message format. Like you said, the branch name might contain numbers. 

What’s the purpose of retrieving a PR number for a coming in the master branch?

Kronopt:
I agree with @sargunv.

You can also get the commit hash with the [git log command](https://git-scm.com/book/en/v2/Git-Basics-Viewing-the-Commit-History#pretty_format) and then do something like [this](https://stackoverflow.com/questions/12214746/find-a-commit-on-github-given-the-commit-hash).

If that can do what you want, I think it would be better and less error prone

Naramsim:
> What’s the purpose of retrieving a PR number for a coming in the master branch?

It was just to let the updater bot post messages about the deployment status. We can get rid of that feature if it seems unstable. I just thought it was quite cool.

Kronopt:
It is cool!

Naramsim:
I still want to have this feature.

Now the PR number is extracted using a regex. Yet, if the number cannot be determined the script can still run well. It will simply not post messages and while opening the pull request towards PokeAPI/api-data it will not reference the merged PR of PokeAPI/pokeapi

Kronopt:
@Naramsim, have you checked [this](https://discuss.circleci.com/t/circle-pull-requests-environment-variable-not-being-set/29883/10) post on the circleci discussion?

Naramsim:
Hi @Kronopt, I've seen it and a similar procedure can be written for Github but it will be a bit different since on Github you cannot filter PRs by branch. Ah, another thing is that $CIRCLE_BRANCH will always be `master` or `staging` since the job runs when the `master`/`staging` branches are pushed to. 

In the end, you could just loop over the PRs and filter out the one with the same HEAD commit SHA. But it's too much of a work, I don't think I will implement it. Feel free to try on your own if you want.

Or you could try to use Github's GraphQL

Naramsim:
Ok, with GraphQL it can be done:

```graphql
query associatedPRs($sha: String, $repo: String!, $owner: String!){
  repository(name: $repo, owner: $owner) {
    commit: object(expression: $sha) {
      ... on Commit {
        associatedPullRequests(first:1){
          edges{
            node{
              number
            }
          }
        }
      }
    }
  }
}
```

Variables:

```json
{
  "sha": "d5b43630fabb61b4c9699eff313592b71edbccba",
  "repo": "pokeapi",
  "owner": "PokeAPI"
}
```

Output:

```json
{
  "data": {
    "repository": {
      "commit": {
        "associatedPullRequests": {
          "edges": [
            {
              "node": {
                "number": 488
              }
            }
          ]
        }
      }
    }
  }
}
```

Kronopt:
The GraphQL solution is cool.
And it's much more reliable than using grep/regex on the commit message.

I think this should be the way to do it

pokeapi-machine-user:
A [PokeAPI/api-data](https://github.com/PokeAPI/api-data) refresh has started. In 45 minutes the staging branch of [PokeAPI/api-data](https://github.com/PokeAPI/api-data/tree/staging) will be pushed with the new generated data. <br><br> The staging branch will be deployed in our [staging environment](https://staging.pokeapi.co/api/v2/) and you will be able to review the entire API. <br><br> A Pull Request ([master](https://github.com/PokeAPI/api-data/tree/master)<-[staging](https://github.com/PokeAPI/api-data/tree/staging)) will be also created at [PokeAPI/api-data](https://github.com/PokeAPI/api-data/pulls) and assigned to the PokeAPI Core team to be reviewed. If approved and merged new data will soon be available worldwide at [pokeapi.co](https://pokeapi.co/api/v2/).

PR #508, opened 2020-07-10
Add dream-world sprites

C-Garza:
Extended the Pokemon `sprites` field by adding the dream world sprites. I also extended the max length on the PokemonSprites model because if more images are added like different version sprites it'll exceed the 500 limit.

### DIFF
--- a/data/v2/build.py
+++ b/data/v2/build.py
@@ -31,7 +31,7 @@ DB_VENDOR = connection.vendor
 
 
 MEDIA_DIR = "/media/sprites/{0}"
-IMAGE_DIR = os.getcwd() + "/data/v2/sprites/"
+IMAGE_DIR = os.getcwd() + "/data/v2/sprites/sprites/"
 RESOURCE_IMAGES = []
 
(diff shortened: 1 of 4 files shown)


Naramsim:
Looks promising! I'll merge and let's see if it deploys to https://staging.pokeapi.co (in 45mins)

Naramsim:
Hmm, there is an issue in our pipeline. We always deploy `master`.

https://github.com/PokeAPI/api-data/blob/183af52fe8b7eac8efaee37c33b5fb929c13982e/updater/cmd.bash#L18

So your changes won't be effective. Let me think about a workaround.

Naramsim:
Ok, the best thing would be to `checkout $CIRCLE_SHA1` inside the `cmd.bash` script. We also need to add `--recurse-submodules` to the cloning phase of `pokeapi` since now we have a submodule.

I'll add right now the `--recurse-submodules` and then merge this `staging` branch to the `master` one and see.

Later I'll add the custom checkout.

PR #590, opened 2021-03-11
Fixed Docker Postgres Config

ZSGaming1998:


Naramsim:
Was it broken?

As far that I know Compose allows both ways.

ZSGaming1998:
i couldn't get it to dock on my instance without these setting using latest
of everything

phalt:
That's a pretty significant syntax change?

phalt:
Looks to be correct though: https://docs.docker.com/compose/environment-variables/#set-environment-variables-in-containers

ZSGaming1998:
yes i agree it took me awhile to figure out what was wrong. until i found the syntax change.

ZSGaming1998:
Idk it your repo that was my solution i could not get it to work im on a new windows install everything is up to date if theres any problems here you guys go keep up the good work!

Naramsim:
Hi @ZSGaming1998,

I checked both for Compose 2 and Compose 3, the syntax we use is still correct.

https://docs.docker.com/compose/compose-file/compose-file-v3/#environment

Can you tell me the version of your `docker-compose`? Is it 3.9?

Naramsim:
Closing due to inactivity and unnecessity

PR #663, opened 2021-10-24
Allow dev deployment

Naramsim:
Allows running both `master` and dev environments.

`docker-compose up` will use `pokeapi/pokeapi:master`
`docker-compose -f docker-compose.yml -f docker-compose-dev.yml up -d` will build the local Dockerfile


pokeapi-machine-user:

PR #702, opened 2022-03-07
Add aiopokeapi to wrappers

beastmatser:
I simply added a reference to the asynchronous wrapper that I have written in Python, I would very much appreciate if it would get added


Naramsim:
Thanks a lot! and good job!

Naramsim (approved):

pokeapi-machine-user:

PR #860, opened 2023-03-26
Add italian translations

AndryCola:
i added full italian translation for:
ability names, evolution trigger prose, generation names, item names, langiage names, location names, move names, pokemon form names, region names and pokemon species names.

i added 2 tables:
gender_names to give a name of pokemon's genders
pokemon form evolution for giving a like to pokemon form and id of the evolution that was missing

i added some evolution trigger for paldean pokemon and i added the the method values in the pokemon evolution files and i fixed some missing data about eevee location


AndryCola:
Hi, how can i fix it?
I don't know what the problem.

PR #880, opened 2023-05-10
Add italian translations

AndryCola:
i added some missed moves italian names


Naramsim:
Thanks for the contribution! 

Could you sort the added entries? Many thanks!

PR #908, opened 2023-07-28
feat: add new endpoint that uses pokemon id to search the evolution c…

DylasX:
Hey,

Working with the endpoint evolution-chain, I saw that here you are using the evolution_chain_id to filter and get the chain evolution so if you only have the pokemon_id and try to get the evolutions with that endpoint will generate a not expected response so that's why I added this new endpoint to send the pokemon_id and get the evolution chain.

Disclaimer: I'm not a python developer so if I need to change something just tell me 

Thanks a lot!


DylasX:
Also please could I get help with the formatter please 🙇🏼‍♂️

PR #909, opened 2023-07-28
feat: change logic on chain-evolution endpoint

DylasX:
Hey,

Working with the endpoint evolution-chain, I saw that here you are using the evolution_chain_id to filter and get the chain evolution so if you only have the pokemon_id and try to get the evolutions with that endpoint will generate a not expected response so that's why I did this proposal of change the chain id to pokemon id.

Thanks a lot!


phalt:
Closing as this is old and I want to tidy things up.

PR #911, opened 2023-08-01
Removed Fake-out from Walking-Wake

linnos:
Removed move ID: 252  from PokemonID: 1009
Fake-Out removed from Pokemon: Walking Wake

Tests ran and passed with: make test

Testing: 
*run local api call for pokemon/walking-wake
*fake-out is not present in learn set


Naramsim:
many thanks!

PR #1010, opened 2024-01-19
Made location_area identifiers actually mean something to a human

DaltonSW:
Added region and location information to the location_area identifier fields so it means something to a human reading looking at them


Naramsim:
Hi, we construct the name starting from the location name. So for the [ID 1](https://pokeapi.co/api/v2/location-area/1/) you can see the `name` property correctly set. 

If we merge this PR the first part of the name would be doubled. Ending up in sinnoh-canalave-city-canalave-city-area.

Nonetheless I think we can change some `identifiers`:

- all `galar-*`. [Example](https://github.com/PokeAPI/pokeapi/pull/1010/files#diff-30477f1a63605c7db1e0dc9fbfad05f1dc5e90e20279b5355cef3d8dfa20c261L734). In the current API the name is rendered as galar-route-1-galar-1. So, of your PR I'd keep the `route` addition. 
- remove all `galar-` prefixes

DaltonSW:
Okay, I see what you mean. Is there any place where this kind of stuff is documented? This lack of specific documentation or any references of how the CSVs interact is part of what I was trying to address with this change

For example, if there was something saying" the location area identifier is used to construct the location endpoint name element", I would be able to more specifically target my testing and verifying

DaltonSW:
How would you feel about an extra, optional column being added to some of the CSVs to act as a human-readable descriptor? Some of the barriers to me adding more info to these files is that I have no idea what the hell I'm looking at without busting out a whiteboard or a script to mentally piece everything together 😅

PR #1028, opened 2024-02-01
Use local image when testing k8s

Naramsim:


pokeapi-machine-user:

PR #1031, opened 2024-02-02
Updating MoveDetailSerializer to change $effect_chance to actual number

### LINKED ISSUE #1024: Replace $effect_chance% with actual value
Part of our API serves text fields containing `$effect_chance%`.

Rendered field: https://pokeapi.co/api/v2/move/fire-punch `.effect_entries[0].effect`

Raw data: https://github.com/PokeAPI/pokeapi/blob/master/data/v2/csv/move_effect_prose.csv#L6

It would be awesome if we could replace the `$effect_chance%` with the actual chance of the move. 

An example for `fire-punch` would be to use the value found in `moves.csv` for the column `effect_chance` (which is 10) and replace it in the effect description.

An even-more-interesting thing would be to create a function that we could use to substitute any `$xxx%` with the relative `xxx` attribute.

SKCwillie:
resolve #1024 

Take effects text from the seralizer and loop over dictionary looking for "effect_chance" when it is found, replace it with move_effect_chance.

I know the build file was mentioned in the comments of the issue but I don't see a way to edit the build that can be dynamically edited later. It definitely could be possible but will have to be done b someone with more knowledge on the build file than me.

To test this, it can be deployed to Staging and the endpoints can be hit to confirm changes.

### DIFF
--- a/pokemon_v2/serializers.py
+++ b/pokemon_v2/serializers.py
@@ -2343,7 +2343,12 @@ class MoveDetailSerializer(serializers.ModelSerializer):
         data = MoveEffectEffectTextSerializer(
             effect_texts, many=True, context=self.context
         ).data
-
+        effect_entries = data[0]
+        for i, k in enumerate(effect_entries):
+            print(i, k)
+            if '$effect_chance%' in effect_entries[k]:
+                data[0][k] = effect_entries[k].replace('$effect_chance', f'{obj.move_effect_chance}')
+    
         return data
 
     def get_effect_change_text(self, obj):
--- a/data/v2/csv/move_effect_prose.csv
+++ b/data/v2/csv/move_effect_prose.csv
@@ -153,9 +153,7 @@ This move cannot be copied by []{move:mimic} or []{move:mirror-move}, nor select
 The user takes damage instead of being healed if the target has []{ability:liquid-ooze}.
 
 []{move:rapid-spin} will remove this effect.
-
-This effect is passed on by []{move:baton-pass}."
-86,9,Does nothing.,"Does nothing.
+7,9,Has a $e.
 
 This move cannot be used while []{move:gravity} is in effect."
 87,9,Disables the target's last used move for 1-8 turns.,"Disables the target's last used move, preventing its use for 4–7 turns, selected at random, or until the target leaves the [field]{mechanic:field}.  If the target hasn't used a move since entering the [field]{mechanic:field}, if it tried to use a move this turn and [failed]{mechanic:failed},  if its last used move has 0 PP remaining, or if it already has a move disabled, this move will fail."


SKCwillie:
Sorry for all the commits. I keep finding issues or small things that can make the code a little cleaner.

GreatNovaDragon:
i've got a question: is that the text that is displayed via the api, or the internally saved text that is changed here? (I dont know how pokeapi works internally)

I dont think the internal text should be changed, only the displayed text.

Naramsim:
@SKCwillie thanks for trying and figuring out how the code works :) 

Could you please fix the linting issues? Running `black .`. Afterwards I can deploy in `staging`

SKCwillie:
@Naramsim checks are passing now!

GreatNovaDragon (approved):

simonorono (changes requested):
This is great for the API, however this won't translate to the GraphQL implementation, for that we would need to do this while building the data.

pokemon_v2/serializers.py
```diff
@@ -2343,6 +2343,12 @@ def get_effect_text(self, obj):
         data = MoveEffectEffectTextSerializer(
             effect_texts, many=True, context=self.context
         ).data
+        effect_entries = data[0]
+        for _i, k in enumerate(effect_entries):
+            if "$effect_chance%" in effect_entries[k]:
+                data[0][k] = effect_entries[k].replace(
+                    "$effect_chance", f"{obj.move_effect_chance}"
+                )
```
simonorono: This code fails for moves without effect entry like dire claw or hard press (both introduced in gen 9).

It could also be written without needing extra or unused variables; here are my suggested changes:

```python
if len(data) > 0:
    for key, value in data[0].items():
        if "$effect_chance%" in value:
            data[0][key] = value.replace(
                "$effect_chance", f"{obj.move_effect_chance}"
            )
```

Naramsim:
In the meantime I deployed the current version to the staging site:

https://staging.pokeapi.co/api/v2/move/fire-punch
https://pokeapi.co/api/v2/move/fire-punch

https://staging.pokeapi.co/api/v2/move/dire-claw
https://pokeapi.co/api/v2/move/dire-claw

But actually what @simonorono says is of great value. I forgot we also have the GQL 😅

simonorono:
I just realized that it's not possible to do this during the build of the database since several moves share the same effect entry, e.g. fire punch, lava plume, inferno, etc. all share the same effect entry.

Naramsim:
> I just realized that it's not possible to do this during the build of the database since several moves share the same effect entry, e.g. fire punch, lava plume, inferno, etc. all share the same effect entry.

So PokeAPI's Postgres has links for different moves to the same effect entry and it's all dynamic. Correct? Yeah, just checked. So the build step is off-limits. 

I see two routes:
- Merging the PR with Simon's fixes and possibly having the GraphQL API only show `$effect_chance%`. 
- Creating and running a tool that beforehand updates the static CSV. This has to be well thought out though. 
- Don't do anything. Just manually correct the effect entries and remember that when new data comes in also the effect description of the move has to change.

simonorono:
Another solution could be having the build step generate effect entries for every different value of `move_effect_chance`

image

Naramsim:
> Another solution could be having the build step generate effect entries for every different value of `move_effect_chance`

I don't follow you very well on this but I'm lacking SQL/Django knowledge at the moment.

Naramsim:
Possible solution for GQL: https://hasura.io/docs/latest/schema/postgres/computed-fields/

pokemon_v2/serializers.py
SKCwillie: I pushed your suggested changes. Definitely a good catch on moves without effect entries

simonorono (approved):

pokeapi-machine-user:

PR #1078, opened 2024-05-06
Query q OAS descriptor

Naramsim:
@Indyandie could you review this PR? Thanks!

PR #1087, opened 2024-05-10
docs(openapi): `q` query string param, fix `id` path parameter, nix refactor

Indyandie:
# OpenAPI Docs

- `q` query string param
- fix `id` path parameter
- nix refactor

Related #1078 #1065

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,21 +1,10 @@
 with import <nixpkgs> { };
-
-let
-  pythonPackages = python3Packages;
-in pkgs.mkShell rec {
-  name = "impurePythonEnv";
+pkgs.mkShell {
+  name = "onix-shellder";
   venvDir = "./.venv";
   buildInputs = [
     python310Packages.python
     python310Packages.venvShellHook
-
-    # python310Packages.coverage
-    # python310Packages.python-mimeparse
-    # python310Packages.python-dateutil
-    # python310Packages.drf-spectacular
-    # python310Packages.djangorestframework
-    # python310Packages.django-redis
-    # python310Packages.django-cors-headers
 
     # Required dependancies 
     black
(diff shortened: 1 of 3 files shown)


Indyandie:
@Naramsim

Naramsim:
Hi, when trying your schema here: https://editor-next.swagger.io/

I don't see the `q` parameter for `/abilities/`, is it missing somewhere? On the other hand, if I check out `/berries/` I have `q` but it's weirdly formatted:
swagger

Also, I don't really like the fat that we are adding other code to every method just for documenting one querystring. Can't we treat `q` like `offset` or `limit`?

Indyandie:
I'll take a look and try to figure how to apply it like `offset` and `limit`. Where are those defined? Sorry i am not too familiar with django... 😅

Indyandie:
@Naramsim I got rid of all that unnecessary code. All the endpoints that use `PokeapiCommonViewset` should pick up the definitions. The formatting is probably that swagger isn't reading the `quoteblock` properly.

Naramsim:
Nice! Thanks a lot! If you have time, would you be able to fix the quoteblock strange formatting? Also, in the console I see other errors and warnings. 

Many thanks!

pokeapi-machine-user:

Indyandie:
> Nice! Thanks a lot! If you have time, would you be able to fix the quoteblock strange formatting? Also, in the console I see other errors and warnings.
> 
> Many thanks!

@Naramsim what command is generating the errors and warnings, can you share them? Or can you create an issue with all the details?

Naramsim:
If you go [here](https://editor-next.swagger.io/) and input your openapi URL and then wait 15 seconds, a popup opens up in the bottom part and you can see something like:



Severity | Line | Code | Message
-- | -- | -- | --
  | 9320 | 7010601 | tags must be an array of Tag Objects
  | 2795 | 10069 | property "example" is deprecated, use "examples" instead

PR #1159, opened 2024-11-01
fix #1067: Icy Snow Vivillon incorrect image

### LINKED ISSUE #1067: Vivillon Icy Snow Image Incorrect
When pulling data for the Vivillon forms, I noticed that the default form image for Vivillon (666) was the **Meadow** form. However, when going into the data call, it was specifying the _666-meadow.png_ image and was not specified as the default form (is_default: false).

When I grabbed the data for the **Icy Snow** form, this did not specify form within the image and just used _666.png_ image. This form was listed as the default (is_default: true). The default _666.png_ image is also the **Meadow** form, which is incorrect for the call being made.

Steps to Reproduce:

1. Access these two API calls:
- https://pokeapi.co/api/v2/pokemon-form/666 (Meadow, default pokemon ID)
- https://pokeapi.co/api/v2/pokemon-form/10086 (Icy Snow, specific form ID)
2. Look into the sprites for each form:
- Meadow Form page will have https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/666-meadow.png
- Icy Snow Form page will have https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/666.png
3. Both of these images are the same, therefore giving the Icy Snow form the incorrect image

Possible Solutions:
- There is an available https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/666-icy-snow.png image to use, it is just not referenced anywhere. This would need to be changed for each kind of image display (HOME, Official, etc.) that is attached to both the pokemon and pokemon-form endpoints.
- The default https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/666.png should be changed the Icy Snow form image. However, this would be a widespread change as there are many images for each kind of image display (HOME, Official, etc.)

rayusagi:
Relates to #1067.

I opted for the solution suggested by @Deleca7755. It seemed like the easiest one and the original intended objective.
The Meadow form is now the default one instead of the Icy Snow form. 
Consequently, _almost_ all existing sprites don't need to be changed. 
There is only the default shiny one that points to the Icy Snow form.
I initiated a PR to fix this aswell on the sprites project: https://github.com/PokeAPI/sprites/pull/157

### DIFF
--- a/data/v2/csv/pokemon_forms.csv
+++ b/data/v2/csv/pokemon_forms.csv
@@ -664,7 +664,7 @@ id,identifier,form_identifier,pokemon_id,introduced_in_version_group_id,is_defau
 663,talonflame,,663,15,1,0,0,1,858
 664,scatterbug-icy-snow,icy-snow,664,15,1,0,0,1,859
 665,spewpa-icy-snow,icy-snow,665,15,1,0,0,1,879
-666,vivillon-meadow,meadow,666,15,0,0,0,7,905
+666,vivillon-meadow,meadow,666,15,1,0,0,7,905
 667,litleo,,667,15,1,0,0,1,919
 668,pyroar,,668,15,1,0,0,1,920
 669,flabebe-red,red,669,15,1,0,0,1,921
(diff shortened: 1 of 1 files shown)


Naramsim (approved):

pokeapi-machine-user:

rayusagi:
Thanks!
However, why did you tag with 'no-deploy'? It's essential to be deployed since https://github.com/PokeAPI/sprites/pull/157 and https://github.com/PokeAPI/sprites/pull/155/ have been merged.

I see the latter PR removed a the `666-meadow.png` sprite. Unfortunately, without this current fix, the API provides a broken link here : https://pokeapi.co/api/v2/pokemon-form/666/ in `sprites.front_default`.

Making the meadow form the default one (what this PR does) should fix the issue and point towards https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/666.png that does exist.

Naramsim:
Because I merged 2 pr and wanted the last one merged to be the only one deployed

rayusagi:
Oh okay, good to know.

PR #1174, opened 2024-12-19
Update docker-compose.yml

Naramsim:


phalt (approved):
A wider gap to give it more time to boot up?

Naramsim:
The problem was that a 1s interval is too frequent, especially for the graphql machine. I saw that pg_isready consumes quite a lot of CPU and was slowing down the entire machine especially during the gql data update procedure.

pokeapi-machine-user:

Issue #1179, opened 2024-12-25
[BUG] Parameters should be case insensitive


pedwards95:
Steps to Reproduce:

1. Submit a request with capitals. eg. https://pokeapi.co/api/v2/pokemon/Iron-Crown
2. Receive no reply (404)

Requests with capitals should still work. Very easy fix just to lowercase parameters.

Naramsim:
Hi! Could you try and implement the fix?

pedwards95:
Yeah I can put together a PR when I get back from vacation.

pedwards95:
Update: After reviewing data, and in relation to asinine choice like the use of hectograms, and the refusal to fix it as seen in this ticket https://github.com/PokeAPI/pokeapi/issues/1182
I will not be fixing this or any of the other issues encountered.

FallenDeity:
Fixed with #1190 and PokeAPI/deploy#50, we can close this now

andre-bacci:
Shouldn't this be closed by now? @Naramsim (don't know who else I can tag)

PR #1190, opened 2025-01-21
feat: Make name lookups in api endpoints case insensitive

FallenDeity:
Solves issue #1179 makes lookup via name case-insensitive, for example 

```py
In [1]: import requests as r

In [2]: x = r.get('http://localhost:8000/api/v2/pokemon/Bulbasaur/')

In [3]: y = r.get('http://localhost:8000/api/v2/pokemon/bulbasaur/')

In [4]: x.status_code, y.status_code
Out[4]: (200, 200)

In [5]: x.json()["name"], y.json()["name"]
Out[5]: ('bulbasaur', 'bulbasaur')
```

Minor changes in lookup logic now uses `[__iexact]`(https://docs.djangoproject.com/en/dev/ref/models/querysets/#iexact) for querying.

Test cases are added for case-insensitive testing for the api:
https://github.com/PokeAPI/pokeapi/blob/63f0e2699b713dd1f832aba7f6c53eb66a6a30c6/pokemon_v2/tests.py#L5695-L5799

@Naramsim feel free to take a look when you have some time

### DIFF
--- a/pokemon_v2/api.py
+++ b/pokemon_v2/api.py
@@ -39,7 +39,8 @@ class NameOrIdRetrieval:
     """
 
     idPattern = re.compile(r"^-?[0-9]+$")
-    namePattern = re.compile(r"^[0-9A-Za-z\-\+]+$")
+    # Allow alphanumeric, hyphen, plus, and space (Space added for test cases using name for lookup, ex: 'base pkm')
+    namePattern = re.compile(r"^[0-9A-Za-z\-\+ ]+$")
 
     def get_queryset(self):
         queryset = super().get_queryset()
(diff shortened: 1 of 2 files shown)


phalt:
This is a cool implementation but it is worth noting that this won't be inherited by the live pokeapi.co instance - the way we build and deploy that generates thousands of static files that we upload, and there is no live server to do the case insensitive look up.

An alternative would be to use Ditto (the tool we use) to crawl each endpoint with all possible case inputs, but that would generate a gigantic number of files and might not be applicable.

We could also update the small app function we use to lower case the input?

FallenDeity:
That works too 🙂, just lower casing the query params should do the trick, afaik all our resource names are also lower case so we should be fine

pokeapi-machine-user:

FallenDeity:
I'll take a look at the repo and make a pr by hopefully tomorrow 🙂

PR #1221, opened 2025-03-09
Added missing flavor text entries

aserra54:
Adds some missing flavor text entries for moves in Pokemon: Legends Arceus and Pokemon Scarlet/Violet. The data was sourced from Bulbapedia.

I'm submitting a small subset of changes to see if they line up with what you expect before making any larger scale changes. Notably, I'm not 100% sure if you want to preserve the line breaks from the original game, or if putting the plain text with no breaks is okay.

The generation and sorting of the file is done automatically via a one-off Python script to ensure the contents are placed in the correct locations. I can share the script if need be.

To help with some context:

```
Move ID 33: Tackle
Move ID 47: Sing
Move ID 421: Shadow Claw
Move ID 827: Dire Claw
Move ID 828: Psyshield Bash

Version ID 24: Pokemon: Legends Arceus
Version ID 25: Pokemon Scarlet & Violet

Language ID 9: English
```


Naramsim:
Hi! This is great! Thanks for the contribution! I'd say that if you have some time and it's feasible/easy to even add in the line-breaks, than that would be perfect. If not, we can go with no line-breaks.

aserra54:
Sure, give me some time to see if I can find a way to make the process as painless as possible on my end, and I will get back to you.

aserra54:
Closing in favor of https://github.com/PokeAPI/pokeapi/pull/1223.

PR #1252, opened 2025-05-31
Feature: Added devcontainer support

### LINKED ISSUE #1251: Feature: Add Dev Container for Easy Development Setup
It will be awesome to have a [dev container](https://code.visualstudio.com/docs/devcontainers/containers) setup that spins up everything - app, DB, cache, GraphQL - so anyone can jump into development without worrying about manual setup.

### Why Dev Containers?
When I first started working on this project, it was pretty Linux-focused, and since I was on Windows, the setup wasn’t exactly easy. Having a dev container would make the setup way easier and smoother for everyone.

If this idea sounds good, I’m happy to start working on it!

PraaneshSelvaraj:
## What’s Added

This PR adds support for [Dev Container](https://code.visualstudio.com/docs/devcontainers/containers). It makes it easy to get started with PokeAPI in a consistent, containerized development environment.

## Highlights

- Added `.devcontainer/devcontainer.json`:
  - Uses `docker-compose.yml` and `docker-compose-dev.yml`
  - Sets up Python, Black formatter, and common ports (8000, 8080, 80, 443)
  - Automatically installs tools and builds the database on startup

## How to Use

1. Open the repo in VS Code
2. Open the Command Palette (`Ctrl+Shift+P` or `Cmd+Shift+P`) and run:  
   `Dev Containers: Reopen in Container`
3. You’re ready to develop!

## Why It’s Useful

- No need to install dependencies manually
- Works on Windows, macOS, and Linux
- Speeds up onboarding for new contributors

Fixes #1251


Naramsim:
Hi, I suggest two things: 

- initially write a simple instraction section in the CONTRIBUTING.md file referencing how to use the devcontainer
- second, to use the docker-compose.override.yml so to pull pokeapi/pokeapi from docker hub without building it locally. And maybe we could not build the DB since it takes a long time.

Naramsim:
in my case the HTTPS port didn't work and the 80 was mounted on http://localhost:35499/api/v2/ . So a bit counterintuitive. 

Secondly, the git inside the container not usable, nor detected by VSCode that prompts me to install git. I think this devcontainers aren't really useful afterall, honestly.

PraaneshSelvaraj:
Hey, I will work on fixing these issues. After that, we can decide whether to keep or remove the devcontainer based on what works best.

PraaneshSelvaraj:
@Naramsim I have fixed the issues and added instructions in the CONTIRUBTING.md. Do check it out and tell me if you face any issues.

.devcontainer/docker-compose.override.yml
Naramsim, line 1: Hi, Can't we use this override https://github.com/PokeAPI/pokeapi/blob/master/docker-compose.override.yml ?

PraaneshSelvaraj: The image from Docker Hub automatically starts the Gunicorn server, but for our Dev Container setup, we actually don't want it running by default. I've overridden the CMD to prevent that, as we use `make` commands to handle server startup.

To keep the main `docker-compose.override.yml` clean for its original purpose, I set up a separate override file just for the Dev Container here: `.devcontainer/docker-compose-override.yml`.

PraaneshSelvaraj: For Dev Container features like [docker-outside-of-docker](https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker) (and others), we typically need a [Debian-based](https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker#:~:text=This%20Feature%20should%20work%20on%20recent%20versions%20of%20Debian/Ubuntu%2Dbased%20distributions%20with%20the%20apt%20package%20manager%20installed.) image for them to work properly. 

Since our current image is built on Alpine Linux, I would suggest creating a separate `Dockerfile` inside the .devcontainers folder. This dedicated `Dockerfile` would then build our Dev Container image using a Debian base. 

What do you think of this?

Naramsim: Hi, honestly I think it's too much work with few benefits :) If you like you could work on other issues. 

Top two would be: helping @Indyandie with the new Astro website, polishing the openapi schema and make it available in the current website

PraaneshSelvaraj: Hey @Naramsim,

Totally get where you're coming from on keeping things simple, and I completely agree that we shouldn't add complexity without good reason.

But, just wanted to quickly mention why I think Dev Containers could actually save us headaches long-term: it means anyone can jump in and contribute instantly, without wrestling with setup. It ensures everyone's development environment is the same, so we avoid  "it works on my machine" issues. Plus, it works well with features like cloud development (like GitHub Codespaces).

It feels like a small upfront effort for a much smoother ride for all contributors, especially new ones.

If you still think this isn't required, I'll totally drop it and focus on other things.

Naramsim: Hi @PraaneshSelvaraj sorry for coming back only now. I'm currently unable to fully review the devcontainer file honestly. Maybe I'll just merge it in and see if issues arise when people start using it. Still unsure. I have limited time currently

PraaneshSelvaraj: Hey @Naramsim , Sounds good.

Issue #639, opened 2021-08-04
Evolution trigger/condition for region-specific evolutions


ratquaza:
Small request, but possibly adding an evolution-trigger or evolution detail that specifies that a certain evolution is unique to a specific region only. For example, Meowth's evolution data lists that they can evolve into either Persian or Perrzerker, but there is nothing that specifies Perrzerker's evolution being locked to Galar only, or that the evolution only works on a Galarian Meowth.

Naramsim:
We can add it to the evolution chain.

What are the exact conditions for both evolutions?

ratquaza:
Meowth must be level 28 and its Galar variant in order to evolve to Perrserker.

joseph-bayer:
Are there any updates to this?

I don't see a way to tell which form Perrserker evolves from in the [evolution chain api call](https://pokeapi.co/api/v2/evolution-chain/22/). Maybe there's an alternate solution I'm not seeing?

Not sure if this is helpful, but here's a list of Pokemon that would benefit from this improvement:
- Perrserker
- Sirfetch'd
- Mr. Rime
- Cursola
- Obstagoon
- Runerigus
- Overqwil
- Sneasler
- Basculegion

dpMelian:
I second this petition. In my app I display a list of the evolution line but I have no way to distinguish between Slowbro and Slowbro Galar's evolution triggers. From what I can see I think the result is in order, i.e. the first object is Slowbro and the second one is Galar's form but maybe this is not the case for every single Pokémon variant. I think adding a `variant` reference inside the `evolution_details` would be really useful

Naramsim:
I'll promote this to a `help wanted` issue and pin it to the top. Maybe some volunteers will show up.

GreatNovaDragon:
Basically any issue here should have the label of "Help wanted", cause the amount of users far eclipses the amount of helpers that actually contribute. 

Data from Galar, Hisui and Paldea basically doesnt exist besides actual stats and names that got extrated from the games.

joereg4:
Would it be something like this?

{
  "evolution_chain": {
    "id": 33,
    "chain": {
      "species": "slowpoke",
      "evolves_to": [
        {
          "species": "slowbro",
          "evolution_details": [
            {
              "trigger": "use_item",
              "item": "water-stone",
              "region_restriction": null,
              "base_form_required": null
            },
            {
              "trigger": "use_item", 
              "item": "galarica-cuff",
              "region_restriction": "galar",
[...]

PR #1299, opened 2025-09-13
Add regional evolution metadata support

joereg4:
- Add region_restriction and base_form_required fields to PokemonEvolution model
- Update PokemonEvolutionSerializer to include new fields in API responses
- Add migration for new fields
- Add implementation plan documentation
- Add data import script for regional evolution data

This addresses GitHub issue #639 by adding support for regional evolution requirements like Galarian Yamask → Runerigus and Galarian Slowpoke → Slowbro.

### DIFF
--- a/import_regional_evolutions.py
+++ b/import_regional_evolutions.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+"""
+Import regional evolution data from CSV into the database
+"""
+import os
+import sys
+import csv
+import django
+
+# Setup Django
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'config.local')
+django.setup()
+
+from pokemon_v2.models import (
+    PokemonEvolution, PokemonSpecies, Region, Item, Location, 
+    EvolutionTrigger, Move, Type
+)
+
+def get_or_create_trigger(name):
+    """Get or create an evolution trigger"""
+    trigger, created = EvolutionTrigger.objects.get_or_create(name=name)
+    return trigger
+
+def get_or_create_item(name):
+    """Get or create an item"""
+    if not name or name.lower() in ['null', '']:
+        return None
+    try:
+        return Item.objects.get(name=name)
+    except Item.DoesNotExist:
+        print(f"⚠️  Item '{name}' not found")
+        return None
+
+def get_or_create_location(name):
+    """Get or create a location"""
+    if not name or name.lower() in ['null', '']:
+        return None
+    try:
+        return Location.objects.get(name=name)
+    except Location.DoesNotExist:
+        print(f"⚠️  Location '{name}' not found")
+        return None
+
+def get_or_create_region(name):
+    """Get or create a region"""
+    if not name or name.lower() in ['null', '']:
+        return None
+    try:
+        return Region.objects.get(name=name)
+    except Region.DoesNotExist:
+        print(f"⚠️  Region '{name}' not found")
+        return None
+
+def get_or_create_species(name):
+    """Get or create a Pokemon species"""
+    if not name or name.lower() in ['null', '']:
+        return None
+    try:
+        return PokemonSpecies.objects.get(name=name)
+    except PokemonSpecies.DoesNotExist:
+        print(f"⚠️  Pokemon species '{name}' not found")
+        return None
+
+def import_regional_evolution_data(csv_file):
+    """Import regional evolution data from CSV"""
+    print(f"Importing regional evolution data from {csv_file}...")
+    
+    imported_count = 0
+    skipped_count = 0
+    
+    with open(csv_file, 'r') as f:
+        reader = csv.DictReader(f)
+        
+        for row_num, row in enumerate(reader, start=2):  # Start at 2 for header
+            try:
+                # Get Pokemon species
+                pokemon_name = row['pokemon']
+                evolves_to_name = row['evolves_to']
+                
+                pokemon = get_or_create_species(pokemon_name)
+                evolves_to = get_or_create_species(evolves_to_name)
+                
+                if not pokemon or not evolves_to:
+                    print(f"⚠️  Skipping row {row_num}: Missing Pokemon species")
+                    skipped_count += 1
+                    continue
+                
+                # Get evolution trigger
+                method = row['method']
+                trigger = get_or_create_trigger(method)
+                
+                # Get other fields
+                region = get_or_create_region(row['region'])
+                item = get_or_create_item(row['item'])
+                location = get_or_create_location(row['location'])
+                base_form = get_or_create_species(row['base_form_required'])
+                
+                # Parse level
+                min_level = None
+                if row['level'] and row['level'].strip():
+                    try:
+                        min_level = int(row['level'])
+                    except ValueError:
+                        print(f"⚠️  Invalid level '{row['level']}' in row {row_num}")
+                
+                # Create or update evolution
+                evolution, created = PokemonEvolution.objects.get_or_create(
+                    evolved_species=evolves_to,
+                    evolution_trigger=trigger,
+                    defaults={
+                        'evolution_item': item,
+                        'min_level': min_level,
+                        'location': location,
+                        'region_restriction': region,
+                        'base_form_required': base_form,
+                    }
+                )
+                
+                if created:
+                    print(f"✅ Created evolution: {pokemon_name} → {evolves_to_name}")
+                    imported_count += 1
+                else:
+                    # Update existing evolution with new fields
+                    evolution.region_restriction = region
+                    evolution.base_form_required = base_form
+                    evolution.save()
+                    print(f"🔄 Updated evolution: {pokemon_name} → {evolves_to_name}")
+                    imported_count += 1
+                    
+            except Exception as e:
+                print(f"❌ Error processing row {row_num}: {e}")
+                skipped_count += 1
+    
+    print(f"\nImport completed!")
+    print(f"✅ Imported/Updated: {imported_count}")
+    print(f"⚠️  Skipped: {skipped_count}")
+
+def main():
+    """Main function"""
+    csv_file = "regional_evolutions.csv"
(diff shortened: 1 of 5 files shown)


pauljoda:
I think this is a no brainer, as currently the api has a huge oversight as far as how this is handled. 

This references a pokemon species for the form foreign key, it's my understanding that the species is the reference to the base forms data, though that object does contain a list of all forms. Due to this needing to reference a specific form, would it make more sense to instead link to the Pokémon id itself, and if species data is needed it can be queried from that Pokémon's species? This ensures the link is to the exact form required for the evolution, and does not require more implied logic to parse out based on the name passed with the species link.

Naramsim:
I'm away for some days and can't review these PRs at the moment. Thanks for submitting it

Naramsim:
Hi! Can you fix the CICD pipeline and continue your work with AI? It's promising.

joereg4:
@Naramsim fixed the CI issues:
- Updated migration dependency to reference correct parent migration
- Applied black formatting to match project standards
- Removed orphaned database table causing test failures

All CI checks are now passing.

joereg4:
@Naramsim Good morning, not sure if there is anything I can do regarding the graphql load timing out in the Container / k8s check.  If there is something i can tweak please let me know:)

joereg4:
@Naramsim Seeing that the loading of graphql is timing out.  Added some debugging to see where it's getting hung up and also increased the timeout in case the CI environment is just slow.

joereg4:
@Naramsim I've been working the load-graphql issues and was able to resolve it.  I see that the last dependabot commit (ee8ef9ffb24a61b6943bf2bb88fb3c9b1d8e5150) failed for the same reason.  debian:buster was end of life and a switch to debian:bullseye in the load-graphql.yaml fixed the issue. 

Once this PR is approved the workflows should pass.  Was able to test with github actions on my fork.

joereg4:
@Naramsim What happens next?  All checks passed:)

Naramsim:
Hi I want to check it out locally! And test it myself and understand the logic. 

Then I'll merge It in the staging branch and deploy it

import_regional_evolutions.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python3
+"""
```
Naramsim: Hi! I don't like this new approach for inserting data. Could you stick to the currently used one? All the build functions are in the `data/v2/build.py` file. The best would be to add the fewest line of code possible.

import_regional_evolutions.py
```diff
@@ -0,0 +1,162 @@
...
+                skipped_count += 1
+
+    print(f"\nImport completed!")
+    print(f"✅ Imported/Updated: {imported_count}")
+    print(f"⚠️  Skipped: {skipped_count}")
+
+
+def main():
+    """Main function"""
+    csv_file = "regional_evolutions.csv"
```
Naramsim: Also, where is this file? I cannot see it committed here.

import_regional_evolutions.py
joereg4: Made the change using the build process

import_regional_evolutions.py
joereg4: Removed this import script and used the build process instead

pauljoda:
@Naramsim realistically, what is the reason to not just have these be separate species for the distinct forms? They have their own types, region, evolution. By not having the pokemon converge on the same species, when they realistically don't share the same data (color, generation, evolution, shape maybe) all of which are not shared with the other "forms" so it seems to be that these pokemon, which already exist as separate pokemon, should instead have their own species. The barrier to determine if a new species should be added should be if any of the foreign keys change for the form, in which case it should be a different species. 

This would organically fix this issue, and prep for whatever wild changes are coming down the line, especially as newer games more and more frequently add new forms/evolutions to pokemon.

Naramsim:
Hi @pauljoda no reason at all. We should get rid of all these species/forms and variety. We just don't have anyone willing to work on a possible schema and converting the existing data to it.

pauljoda:
I can take a crack at it, for reference I'm using the data for an iOS app, and actually used the Postgres dump to load it into a sqlite db to ship with the app, it quickly got way too network heavy to use the api calls. 

This would be a pretty big refactor, are you saying to flatten all forms, species, and varieties into the Pokémon table itself? I think there is some benefit to having a way for forms to share a common data with the root Pokémon, providing the form is mostly cosmetic. I do think that regional forms should be completely separate Pokémon. 

Perhaps, the structure of Pokémon/Species could still be used, but with the purpose now being Form/Pokemon, where the species table now takes the role of the base pokemon, and the pokemon table now takes the role of the differentiating factors per form. 

Any thoughts or suggestions before I work on that? I think that would be a far more future proof and simple method than adding a new table here, though this pr does solve the problem with the current API so might be worth pulling into for a quick fix and this change come in a future version.

joereg4:
@pauljoda @Naramsim  I understand there are changes people are looking to make that are a heavy lift.  Can we wrap this one up while other work is being done?  I would like to see how I can contribute to these other efforts but see one newer PR on top of this one.  I also see a lot that are years old.  Think we need to close these out and then come up with a strategy that works for the future changes.

pauljoda:
@joereg4 I'm sure it's fine, I think large changes like this should be for v3. I'll move this discussion to a thread so this can remain related for this change on v2

pauljoda:
> @joereg4 I'm sure it's fine, I think large changes like this should be for v3. I'll move this discussion to a thread so this can remain related for this change on v2

https://github.com/PokeAPI/pokeapi/discussions/1305

data/v2/csv/pokemon_evolution.csv
```diff
@@ -1,508 +1,508 @@
...
-503,975,3,885,,,,,,,,,,,,,,,0,0
-504,979,10,,,,,,,,,,,,,,,,0,0
-505,980,1,,20,,,,,,,,,,,,,,0,0
-506,981,1,,,,,,,888,,,,,,,,,0,0
-507,982,1,,,,,,,887,,,,,,,,,0,0
-508,983,10,,,,,,,,,,,,,,,,0,0
-509,997,1,,35,,,,,,,,,,,,,,0,0
-510,998,1,,54,,,,,,,,,,,,,,0,0
-511,1000,10,,,,,,,,,,,,,,,,0,0
+id,evolved_species_id,evolution_trigger_id,trigger_item_id,minimum_level,gender_id,location_id,held_item_id,time_of_day,known_move_id,known_move_type_id,minimum_happiness,minimum_beauty,minimum_affection,relative_physical_stats,party_species_id,party_type_id,trade_species_id,needs_overworld_rain,turn_upside_down,region_restriction,base_form_required
```
Naramsim: Great job! We are close to merging this in! But beforehand I'd like to clean a bit the names used. Can we change `region_restriction` to `region_id` and `base_form_required` to `base_form` or `base_form_id` (if it's an ID)? So that the colum names are inline with the existing fields, such as `region_id`.

pokemon_v2/models.py
```diff
@@ -1694,6 +1694,24 @@ class PokemonEvolution(HasEvolutionTrigger, HasGender):
 
     turn_upside_down = models.BooleanField(default=False)
 
+    # Regional evolution fields
+    region_restriction = models.ForeignKey(
```
Naramsim: If you agree I'd go with the same names here.

data/v2/csv/pokemon_evolution.csv
joereg4: Done, pushed the change

pokemon_v2/models.py
joereg4: Done, pushed the change

joereg4:
@Naramsim Made the changes requested, please take a look:)

Naramsim (approved):

Naramsim:
Great! Thanks for sticking with us on this PR :)

Naramsim:
Could you also update the docs at pokeapi.co repo?

pokeapi-machine-user:

pokeapi-machine-user:

data/v2/csv/pokemon_evolution.csv
```diff
@@ -1,508 +1,508 @@
...
+4,6,1,,36,,,,,,,,,,,,,,0,0,,
+5,8,1,,16,,,,,,,,,,,,,,0,0,,
+6,9,1,,36,,,,,,,,,,,,,,0,0,,
+7,11,1,,7,,,,,,,,,,,,,,0,0,,
+8,12,1,,10,,,,,,,,,,,,,,0,0,,
+9,14,1,,7,,,,,,,,,,,,,,0,0,,
+10,15,1,,10,,,,,,,,,,,,,,0,0,,
+11,17,1,,18,,,,,,,,,,,,,,0,0,,
+12,18,1,,36,,,,,,,,,,,,,,0,0,,
+13,20,1,,20,,,,,,,,,,,,,,0,0,7,19
```
Naramsim, line 14: Wait a minute, I did a check and this data doesn't seem correct. If I browse to https://staging.pokeapi.co/api/v2/pokemon-species/20 (Raticate) it's linked to the evolution chain https://staging.pokeapi.co/api/v2/evolution-chain/7/. But that became wrong with the new information. Because Rattata to become Raticate doens' need to be in Alola region. I think you need to revise a bit this CSV file.

Naramsim: @joereg4 could you take a look at the above?

joereg4: The evolution data for Rattata → Raticate was incorrectly restricted to the Alola region only, preventing the evolution from occurring in other regions. This was identified by @Naramsim in the review of the evolution data.

Issue Details:

- Evolution entries had region_id=7 (Alola) and base_form_id=19 (Rattata)
- This incorrectly suggested that Rattata could only evolve into Raticate in Alola
- The evolution should be available in all regions, not just Alola

#1308

PR #1381, opened 2026-01-23
Test encounters endpoint

Naramsim:


phalt:
Can you describe what you're trying to do here?

Naramsim:
We had an issue in the past where sqlite built pokeapi servers couldn't serve the encounters endpoint. Later it was fixed. I'm trying to test the encounters endpoint.

programgames:
Endpoint : https://pokeapi.co/api/v2/encounter-method/walk
and https://pokeapi.co/api/v2/encounter-method/5 
both working, can we close the issue @Naramsim ?

Naramsim:
No, this is not an issue. It's a change for querying the `:pokemon/encounters` endpoint during the CICD pipeline. But I didn't have time to debug why it fails.

PR #1391, opened 2026-02-03
Add UV for project management, Ruff for formatting.

phalt:
This is the first part of https://github.com/PokeAPI/pokeapi/discussions/1390

# UV package manager

[uv](https://docs.astral.sh/uv/) is fast becoming the standard for managing an entire Python project. It handles package management, environment management, tool management etc. It's great.

The first commit introduces it. We also add `pyproject.toml` which is the new [project dependency standard](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) by Python. It replaces `requirements.txt` and other configurations.

# Ruff auto formatting

[Ruff](https://docs.astral.sh/ruff/) is a Python code formatter. It is super fast, and like `uv` becoming the standard.

The second commit is the "auto" formatted changes - there were very few.

# Ruff manual formatting

This is the big commit. I had to manually go and update all the poorly formatted code in the entire project. Read the commit to see the summary of the changes.

# Docker changes

I'm not the best at Docker, but I finally got it working. Changes are mostly around using `uv` properly.

# postgres job

This job used to run in Docker. For some reason it now hits memory limits. I've decided to just shift this to use "bare metal" and achieve the same result.

## AI support

For the very very large files, I made 3-4 changes, then asked a coding agent to "copy the changes I did for the rest of the file to fix the import errors". Then I verified it by manually reading the diff

## Proof it works

- Formatting passes
- Tests pass
- Serve can be ran locally
- `make build-all` works still
- All the docker commands work.


phalt:
So, turns out we're hitting a memory limit on the github action when compiling the docker images, so I am going to figure out what to do there.

phalt:
Weirdly [this job](https://github.com/PokeAPI/pokeapi/actions/runs/21634250446/job/62355061677?pr=1391) hung forever waiting for something to happen.

phalt:
I think perhaps the `k8s` job isn't always supposed to run like this? cc @Naramsim

Naramsim:
hi! Give me some time to take a look into this. I wasn't expeting a overhaul. I'll have to look into these new tools and confirm they are become the standard. 

Regarding the GH Action, I'll try to look into it. The k8s job probably is failing because you didn't update the k8s configuration files. I'll take a look.

Secondly, the Postgres job is useful for, as you said, testing (and documenting) how to dump and load the database. I think every functionality of PokeAPI should be tested and having some test files that document how to actually do something is the best.

phalt:
I've not set up and used k8s myself before so testing and debugging is slow going.

RE: Postgres job. Does it have to be done via docker? If so, I think porting it to circle CI so I can provision a bigger runner is the best course? It keep hitting memory limits. GitHub runner options are limited.

Naramsim:
No need to run it on Docker

phalt:
Postgres job rewritten to use bare metal instead of docker, now works :)

phalt:
@Naramsim  All tests passing now. Reorganised the commits to be distinct. Best reviewed commit by commit.

Makefile
```diff
@@ -4,66 +4,59 @@ docker_config = --settings=config.docker-compose
...
 
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?# .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 install:  # Install base requirements to run project
-	$(PIP) install -r requirements.txt
-
-dev-install:  # Install developer requirements + base requirements
-	$(PIP) install -r test-requirements.txt
+	uv sync
```
Naramsim, line 14: I'd like a target that checks if `uv` is installed and having it running before the other targets that requires it. 

```Makefile
check-uv:
	command -v uv &>/dev/null || { echo
 '`uv` not detected'; }

setup: check-uv  # Set up the project database
	uv run manage.py migrate ${local_config}

File: Makefile [hunk 2]
```diff
@@ -4,66 +4,59 @@ docker_config = --settings=config.docker-compose
...
 openapi-generate:
-	$(PYTHON) manage.py spectacular --color --file openapi.yml ${local_config}
+	uv run manage.py spectacular --color --file openapi.yml ${local_config}
 
 docker-up:  # (Docker) Create services/volumes/networks
 	docker compose up -d
 
 docker-migrate:  # (Docker) Run any pending migrations
-	docker compose exec -T app python manage.py migrate ${docker_config}
+	docker compose --verbose exec -T app uv run manage.py migrate ${docker_config}
```
Naramsim, line 50: We can remove the `--verbose`

pyproject.toml
```diff
@@ -0,0 +1,35 @@
+[project]
+name = "pokeapi"
+version = "0.1.0"
```
Naramsim, line 3: Let's go with `3.0.0`. When this will be merged I'll draft a new release.

pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+description = "The Pokemon API"
+readme = "README.md"
+requires-python = ">=3.14"
```
Naramsim, line 6: Is it true? Can't python 3.10 be used?

pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+    "coverage==7.13.1",
+    "django==5.2.10",
+    "django-cachalot==2.8.0",
+    "django-cors-headers==4.9.0",
+    "django-redis==6.0.0",
+    "djangorestframework==3.16.1",
+    "drf-spectacular==0.29.0",
+    "gunicorn==23.0.0",
+    "psycopg==3.3.2",
+    "psycopg2-binary>=2.9.11",
```
Naramsim, line 17: We don't use psycopg2 anymore. Can you try removing it?

pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+    "ruff>=0.14.14",
+    "ty>=0.0.14",
+]
+
+[tool.ruff]
+line-length = 120
```
Naramsim, line 23: Honestly I don't like having lines with a maximum length. Nowadays all editors split and wrap the lines automatically. I'd remove the property or set it to a greater value.

README.md
```diff
@@ -37,7 +37,9 @@ A RESTful API for Pokémon - [pokeapi.co](https://pokeapi.co)
 
 - Download this source code into a working directory, be sure to use the flag `--recurse-submodules` to clone also our submodules.
 
-- Install the requirements using pip:
+- Install [uv](https://docs.astral.sh/uv/getting-started/installation/) for Python environment management.
```
Naramsim, line 40: I don't paricularily like adding new toold to projects. I don't see the benefits of using `uv` only the downside of forcing everyone to install it on their machines. It's true that it's a project with many stars and maybe a good community but for pokeapi I don't see any added benefits. If you @phalt or other contributors strongly advocate for using it, I'm ok with it, but if I could decide I wouldn't add it. 

My argument is: Python officially already ships with uv-like features. It has `venv`, can install a `pyprojet.toml`, etc. `uv` does it faster from my understanding.

Naramsim: One very good added benefit of `uv` is having a `lock` file.

Resources/docker/app/Dockerfile
```diff
@@ -19,15 +29,23 @@ ENV DJANGO_SETTINGS_MODULE='config.docker-compose'
...
 
+# Copy application code
 ADD . /code/
 
 RUN addgroup -g 1000 -S pokeapi && \
     adduser -u 1000 -S pokeapi -G pokeapi
 USER pokeapi
 
-CMD ["gunicorn", "config.wsgi:application", "-c", "gunicorn.conf.py"]
+CMD ["uv", "run", "gunicorn", "config.wsgi:application", "-c", "gunicorn.conf.py"]
```
Naramsim, line 51: Is `uv` also necessary for running the app? Why can't I just use `gunicorn`?

Resources/docker/app/Dockerfile
```diff
@@ -19,15 +29,23 @@ ENV DJANGO_SETTINGS_MODULE='config.docker-compose'
...
-COPY --from=builder /usr/local/bin /usr/local/bin
+# Copy UV and virtual environment from builder
+COPY --from=builder /root/.local /root/.local
+COPY --from=builder /code/.venv /code/.venv
+
+# Add UV to PATH
+ENV PATH="/root/.local/bin:$PATH"
+
+# Install runtime dependencies
+RUN apk add --no-cache postgresql-libs libstdc++
```
Naramsim, line 42: Are we sure we need this two packages? We didn't need them before. Only for building the app at line 9

pokemon_v2/tests.py
```diff
@@ -1,7 +1,157 @@
 import json
+
 from rest_framework import status
 from rest_framework.test import APITestCase
-from pokemon_v2.models import *
+
```
Naramsim, line 6: Is it bad to `import *`?

Naramsim:
Can I chime in @sargunv @cmmartti @Indyandie especially for whether to adopt the _astral-sh_ ecosystem or not.

sargunv:
Yup, can confirm these are the new standard and for good reason.

I use them all at work and they're great.

Resources/docker/app/Dockerfile
sargunv: This way it'll use the venv to run, without any activate step

pyproject.toml
sargunv: 80-90 characters is pretty standard in the python world. This is extending it from that default to 120 (which, I'd prefer to keep the default for readability, but 120 isn't the worst)

editor-wrapped lines are quite hard to read.

README.md
sargunv: re: "forcing everyone to install it on their machines"

I'd recommend actually configuring the core tools via [`mise`](https://mise.jdx.dev). That way, any contributor only needs mise, and any other tools (uv, python, docker, anything) is managed + locked by mise in local dev, in CI, in devcontainers, AI agent workspaces, everywhere in a reproducible way.

pokemon_v2/tests.py
phalt: Yes, it is implicit, and not recommended in Python.

pyproject.toml
phalt: Can be, I'd prefer to be on the latest version - uv will "just work" for this.

pyproject.toml
phalt: Can do but it's not necessary - this is if we are publishing the project as a package, which we're not.

Makefile
phalt: That's a good idea!

pyproject.toml
phalt: OO this might be something I missed, good spot.

README.md
phalt: Uv is becoming an industry standard, I use it at my company (3,000+ python engineers, one of the biggest Django projects in the world, we even have a few Python and Django core devs here).

Why uv?

- It manages python versions you don't need to install and manage multiple versions yourself.
- It wraps `virtualenv` and eliminates the need to understand it, reduces barrier to entry.
- It has lock files to enforce specific versions - previously pip didn't fully support this.
- It helps you run tools easily (in a future PR I am going to add `uv run pokeapi build_data` command to eliminate all the manual `import data; data.build` stuff we have).
- It is fast.

For years PokeAPI has been a good example of a great API project, and it has lagged behind in being a great example of a Python project, and it could be one with a bit of TLC - which is what my aim is this year.

I think providing up to date tools is a win win for us (easier to maintain) and also people browsing our code (they see what is standard).

Resources/docker/app/Dockerfile
phalt: Will try and drop them and see if it works

Resources/docker/app/Dockerfile
Naramsim: Ok, `uv` [should forward](https://docs.astral.sh/uv/concepts/projects/run/#signal-handling) `gunicorn`-required signals so we should be good. I just don't like having a `uv` process in the container that doesn't do anything, I'd rather activate the environment and only spawn `gunicorn`. Feel free to change or the resolve the conversation.

README.md
Naramsim: Ok, I'll trust you with this one.

---

@sargunv  I wouldn't go with `mise` since it's developed by only a single developer and I fear that in future the project might die, as `react-static` did (the JS frontend we use).

pyproject.toml
Naramsim: Because 3.10 hasn't reached EOL yet. 

Two question: 

- if I'm on a system that doesn't ship python 3.14 on its package manager yet, as Ubuntu 24 or Debian 13, `uv` will compile the source and install it?
- If on my Debian 11 I have python 3.9, `uv` will install python 3.14 seamlessly and use it to run pokeapi?

README.md
sargunv: Fair enough.

This project's dev env is pretty simple, so `mise` value is relatively small; limited to centralizing / documenting dev env in a single place. But contributing.md and gh actions is only two places, not that bad.

config/docker-compose.py
```diff
@@ -1,6 +1,7 @@
 # Docker settings
 import os
-from .settings import *
+
+from .settings import *  # noqa
```
FallenDeity, line 4: might be nitpicky but it might be better to mention `F403`? https://docs.astral.sh/ruff/rules/blanket-noqa/ (same for all the following blanket suppressions)

Makefile
FallenDeity, line 14: This would also update the lock file if we run it without `--locked` is that intended? or do we want dependencies to stay stable for most part?

pyproject.toml
FallenDeity: also just chiming in perhaps it would be a good idea to segregate the dev tooling/dev dependencies like ruff, ty etc and the primary dependencies https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups

Makefile
phalt: It should run with `--locked` you're right.

phalt:
Had a pretty rough flu and sinusitis for two weeks - I'll get back to this when I am recovered.

FallenDeity:
> Had a pretty rough flu and sinusitis for two weeks - I'll get back to this when I am recovered.

Sorry to hear that, get well soon

programgames:
@phalt Hello, pretty hyped with this tool, any news ?

Naramsim:
The work is done here with smaller changes: https://github.com/PokeAPI/pokeapi/pull/1553 @programgames

phalt:
Alessandro beat me to it! I think this PR can stay open and be put into draft and kept around to help people doing the smaller bits of work.

PR #1437, opened 2026-03-12
Add CSV identifier validation with comprehensive error reporting

LoicOuth:
## Add CSV identifier validation to enforce ASCII slug format

Issue #1436

### Context

The API currently returns `400 Bad Request` when accessing resources with Unicode characters or special characters in their identifiers (e.g., `/api/v2/item/kofu's-wallet`). After discussion in #1436, we agreed that resource identifiers should be treated as **URL-safe ASCII slugs** for consistency across the API.

### Approach

Instead of modifying the API to accept Unicode characters, this PR adds **preventive validation at the data source level** (CSV files). This approach:

- ✅ **Validates early**: Catches invalid identifiers before data is loaded into the database
- ✅ **Prevents future issues**: CI will block PRs with invalid identifiers
- ✅ **Maintains consistency**: Ensures all resources follow the same naming convention
- ✅ **Simple to maintain**: No complex regex or Unicode handling in API code

### Implementation

Added `CSVResourceNameValidationTestCase` in `pokemon_v2/test_models.py`:
- Validates **50 CSV files** with identifier columns
- Enforces pattern: `^[a-z0-9-]+$` (lowercase letters, numbers, hyphens only)
- Reports both missing files and invalid identifiers
- Runs automatically in CircleCI via `make test`

### Test Results

**Expected behavior**: This test **will fail** on master, which is intentional. It identifies 16 invalid identifiers that need normalization : check PR #1438 

### Question for Maintainers

⚠️ **Empty identifiers**: Currently, the test **skips empty identifiers** (e.g., `location_areas.csv` has many rows with empty identifier fields). 

Should empty identifiers be:
- **Option A**: Allowed (current behavior - skip validation)
- **Option B**: Rejected (fail the test if identifier is empty)

Example from `location_areas.csv`:
```csv
id,location_id,game_index,identifier
1,1,1,
2,2,2,
3,3,3,

### CODE REVIEW

File: pokemon_v2/test_models.py [hunk 1]
```diff
@@ -9,3 +12,235 @@ def setUp(self):
...
+        ("pokemon_forms.csv", "identifier"),
+        ("pokemon_habitats.csv", "identifier"),
+        ("pokemon_move_methods.csv", "identifier"),
+        ("pokemon_shapes.csv", "identifier"),
+        ("pokemon_species.csv", "identifier"),
+        ("regions.csv", "identifier"),
+        ("stats.csv", "identifier"),
+        ("types.csv", "identifier"),
+        ("versions.csv", "identifier"),
+        ("version_groups.csv", "identifier"),
```
jemarq04: These all have the same name for the column: `identifier`. Maybe just have this be a list of strings (the base file names) and update things to match?

pokemon_v2/test_models.py
```diff
@@ -9,3 +12,235 @@ def setUp(self):
...
+    ]
+
+    def get_csv_path(self, filename):
+        """Get the absolute path to a CSV file in data/v2/csv/"""
+        from django.conf import settings
```
jemarq04: Why import in this function every time it's called instead of at the top of the file?

Also, considering the function can be put in one line and still be easily read I think maybe we can just retrieve the path directly in the test method.

jemarq04 (changes requested):
The tests are good and well documented - I've left a couple comments. I think it's likely best to wait until the fixes in #1438 are merged before we approve this, to make sure that these new tests would pass with the fixes.

pokemon_v2/test_models.py
LoicOuth: Yeah, maybe it's better. When I started create the test, I didn't know if all the name had the same column name. That's why I did it this way. I will make the modification.

pokemon_v2/test_models.py
LoicOuth: Thanks for the feedback! You're absolutely right about both points. For the import, I apologize - I should have placed it at the top of the file. I will make the changes as you suggested ;).

pokemon_v2/test_models.py
LoicOuth: Change is done ;)

pokemon_v2/test_models.py
LoicOuth: Change is done ;)

jemarq04 (approved):
Great, thanks! Like I mentioned before, I think we'll wait to merge this until #1438 is fully merged, but this will be a good test for future contributions.

pokemon_v2/test_models.py
```diff
@@ -9,3 +13,225 @@ def setUp(self):
...
+
+    This test validates the data source (CSV files) before it's loaded into the database.
+    """
+
+    # Pattern for valid resource identifiers: lowercase letters, numbers, and hyphens only
+    VALID_IDENTIFIER_PATTERN = re.compile(r"^[a-z0-9-]+$")
+
+    # CSV files with 'identifier' column to validate
+    CSV_FILES_TO_VALIDATE = [
+        "abilities.csv",
```
Naramsim: Hi, I prefer an approch where all files are scanned and if a column `identifier` is found, then, we check it.

pokemon_v2/test_models.py
```diff
@@ -9,3 +13,225 @@ def setUp(self):
...
+                            {
+                                "file": filename,
+                                "row": "N/A",
+                                "id": "N/A",
+                                "identifier": "Column 'identifier' not found",
+                            }
+                        )
+                        continue
+
+                    for row_num, row in enumerate(reader, start=2):
```
Naramsim, line 62: Why `start=2`?

Naramsim (changes requested):
Hi, I'd like also an enforcing while building the data. 

Maybe you could take a look here: https://github.com/PokeAPI/pokeapi/blob/905e74b3d0bceba6729579eed0ad5650046b8857/pokemon_v2/models.py#L379

And enforce that the `name` always follows the pattern. A model change is required and probably a migration as well.

pokemon_v2/test_models.py
jemarq04: To skip checking the header line

Naramsim: But then it should be `1`, no? the header should be at position `0`

jemarq04: Ah, I had assumed that it counted differently with this tool. I'll check - I should have been more careful since this is likely all AI

Naramsim: https://docs.python.org/3/library/functions.html#enumerate

jemarq04: Oh wow even more embarrassing, that's core python. I should've noticed lol. Either way, it looks like this doesn't do what I assumed from the beginning - the `start` keyword changes what value `row_num` would begin at, but wouldn't change the value in `row`. I believe this is what helps it accurately report the row number in the file, since the reader may start at `row_num=1`.

Naramsim:
@LoicOuth hi, checking if you were willing to continue the work on this PR.

LoicOuth:
> @LoicOuth hi, checking if you were willing to continue the work on this PR.

Sure, no problem. I'd just like to clarify a couple of things to make sure I know exactly what I need to implement.
- Do I need to add a validation check in the model and make sure to verify that the data matches before inserting it into the database? So my test that checks the CSV files becomes unnecessary, right?
https://github.com/PokeAPI/pokeapi/pull/1437#pullrequestreview-3948428274

- So the automatic file search is no longer required?
https://github.com/PokeAPI/pokeapi/pull/1437#discussion_r2934936156

Or you wan't to keep both maybe, csv checking and validation in the model ?

Thanks for your reply. I won’t have much time right now, but by next week I should be able to work on it.

Naramsim:
Let's keep both, in the model and as a test. Thanks!

jemarq04:
@Naramsim Do you know when the validators are checked? I ran a local test with the updates to this PR (changed `canalave-city` in `locations.csv` to `canalave'-city` just to see) and the validator did not throw any errors when running `make build-db` but the tests were able to catch it.

Naramsim:
Not really, I'll check the docs when I have some time

Naramsim:
I don't have much time, so I'll paste what SO is telling me:

> If you are creating objects via the ORM (e.g., in a script or shell), you must call full_clean() manually before saving.

We need to invoke a `.full_clean()` methos somewhere but I'll let @LoicOuth find exatly where and how.

jemarq04:
I've looked into this a bit more and it looks like it may be more complicated than expected. The `full_clean()` method needs to be called on each of the instances of the django model classes (i.e. `Language`, `PokemonSpecies`, etc.). The objects are created [here](https://github.com/PokeAPI/pokeapi/blob/8d631c17448d88fe96b8a8893c199dcd9203a1a9/data/v2/build.py#L117-L128). Using `bulk_create()` bypasses any validation checks to make the objects quickly and efficiently, which is why it'll build without issue.

Adding in validation during the build process would likely slow it down significantly as these things may need to be routed to some sort of serializer or other class to handle validation and bulk creation properly. A naive approach to just validate the objects created from bulk create is failing, and some searching online may suggest that this may not be the best approach.

Personally, I think that we should avoid trying to implement a validator and just leave it as a check in the tests. We'll still get an error if an identifier fails this check, but without compromising on how quickly the database will build. @Naramsim Thoughts?

Naramsim:
Yeah let's go without validators then!

Naramsim:
Can you remove the validator @LoicOuth ?

LoicOuth:
> Can you remove the validator @LoicOuth ?

Yeah no problem

jemarq04:
@Naramsim I think the PR is looking good - do you think we're good to merge this in?

pokeapi-machine-user:

pokeapi-machine-user:

PR #1446, opened 2026-03-23
Add XD Encounters, Encounter Conditions, Locations, Location Areas, Encounter Methods, and a few more things

notblisy:
The goal of this PR is to add all the encounter locations, conditions, and items for Pokemon obtainable in Pokemon XD: Gale of Darkness. 

This is a fair bit larger than Pokemon Colosseum, as there's 82 Shadow Pokemon in this game, plus a few extra gift Pokemon. 

I sourced my data from:

- https://bulbapedia.bulbagarden.net/wiki/List_of_Shadow_Pok%C3%A9mon#List_of_Shadow_Pok%C3%A9mon_in_Pok%C3%A9mon_XD:_Gale_of_Darkness for shadows

- https://bulbapedia.bulbagarden.net/wiki/Gateon_Port#Trainers for dragonites condition, at the bottom with mirror b

- https://bulbapedia.bulbagarden.net/wiki/Duking for trades

- https://bulbapedia.bulbagarden.net/wiki/Gift_Pok%C3%A9mon#XD:_Gale_of_Darkness - for the gift Pokemon

For the encounters, I added the usual:

- Locations
- Location_Areas
- Encounter_Slots
- Encounters
- Pokemon_items

In addition to all the relevant prose for these. 

On top of that, I've added a new encounter method:

- Pokespot

This is a unique way to catch wild Pokemon in XD, where you leave a snack out for a Pokemon, and one of 3 different Pokemon can come to eat it. Then you get to do a wild battle and capture. 

During my time modeling XD, I noticed that while there IS an NPC-Trade option in encounter_methods.csv, this doesn't allow us to note WHICH Pokemon the player would have to trade. So, I added a new encounter condition called:

- trade

Then, in encounter_condition_values.csv I would describe both the NPCs name and Pokemon to trade them. 

An example is that for a larvitar, duking wants a wooper. So I have the encounter condition value id set to:

- 112,15,trade-duking-wooper,0

Encounter conditions seem to be rarely used so far in the API, but I think they are a great way to add specificity to encounters like trades or gifts outside of just locations. I hope this was okay to add!


notblisy:
I think that for encounter condition value map I might have put the wrong value? Is 0
Or 1 what it needs to be for the condition to be true?

Naramsim:
Hi! Thanks for the PR! I see where you want to go but I don't think adding a new encounter method is a good long term solution. We should expand the current `npc-trade` encounter method to take into account other details, such as which pokemon to trade in and to which npc. It's fine to model it first and then add the data in later. 

Let's take an example. https://www.serebii.net/sunmoon/ingametrades.shtml A machop that is traded for a Spearow in Alola. We have this info here: https://pokeapi.co/api/v2/pokemon/66/encounters (search for `npc-trade`):

```json
encounter_details: [
  {
    chance: 100,
    condition_values: [ ],
    max_level: 9,
    method: {
      name: "npc-trade",
      url: "https://pokeapi.co/api/v2/encounter-method/36/"
    },
    min_level: 9
  }
  ],
max_chance: 100,
version: {
  name: "sun",
  url: "https://pokeapi.co/api/v2/version/27/"
}
```

We should expand the details in the `condition_values` list, I think. So it's better to use the encounter method `npc-trade`, add new encounter conditions `trade-in-pokemon` and `trade-to-trainer`. Then add the _encounter_condition_values_. It's a huge job. 

Or, we could re-think the mechanism for trading, trading is not really _encountering a pokemon_. 

I think these are the two options we have:
- do as we always did, using the `npc-trade` and not having the required conditions such as the trader and the pokemon to trade
- implement and discuss together the best logic expanding from my thoughts above.

Naramsim:
> I think that for encounter condition value map I might have put the wrong value? Is 0
> Or 1 what it needs to be for the condition to be true?

It should be `1`

notblisy:
> Hi! Thanks for the PR! I see where you want to go but I don't think adding a new encounter method is a good long term solution. We should expand the current `npc-trade` encounter method to take into account other details, such as which pokemon to trade in and to which npc. It's fine to model it first and then add the data in later.
> 
> Let's take an example. https://www.serebii.net/sunmoon/ingametrades.shtml A machop that is traded for a Spearow in Alola. We have this info here: https://pokeapi.co/api/v2/pokemon/66/encounters (search for `npc-trade`):
> 
> ```json
> encounter_details: [
>   {
>     chance: 100,
>     condition_values: [ ],
>     max_level: 9,
>     method: {
>       name: "npc-trade",
>       url: "https://pokeapi.co/api/v2/encounter-method/36/"
>     },
>     min_level: 9
>   }
>   ],
> max_chance: 100,
> version: {
>   name: "sun",
>   url: "https://pokeapi.co/api/v2/version/27/"
> }
> ```
> 
> We should expand the details in the `condition_values` list, I think. So it's better to use the encounter method `npc-trade`, add new encounter conditions `trade-in-pokemon` and `trade-to-trainer`. Then add the _encounter_condition_values_. It's a huge job.
> 
> Or, we could re-think the mechanism for trading, trading is not really _encountering a pokemon_.
> 
> I think these are the two options we have:
> 
>     * do as we always did, using the `npc-trade` and not having the required conditions such as the trader and the pokemon to trade
> 
>     * implement and discuss together the best logic expanding from my thoughts above.

Hi, maybe I am misunderstanding, but is this not what I did? The new encounter method I added is for pokespot pokemon, not trades.

For the trades I added encounter a new encounter condition, and then 4 new values which describe which trainer to trade with and which pokemon to send them. 

I think leaving NPC-Trade as an encounter method is fine. Just append them with encounter conditions to let the players know what Pokemon they should have on hand for a trade, and which NPC to speak to if they have a name.

We have the mechanism to add such details, in my opinion it's useful information to have, and encounter_conditions are the perfect way to support this.

I don't mind going back and documenting which NPCs and trainers in conditions for a separate PR. I have to do it later anyway, there's a lot of trades missing. I planned on making a few issues today for missing encounters.

Also, I don't think `trade-in-pokemon`, as in, trading in from another game or transferring from an older game is within scope of this idea. I was just looking at encounter conditions for the NPC-Trade encounter methods.

jemarq04:
I think adding an encounter condition for the Pokemon needed to trade is not a bad idea, but if a change like that is done for these encounters then a new PR should definitely be made for previous ones. While Duking is an easy enough character to recognize, not every trainer with trades in-game will have a name at all so I think a condition like that is not easily scaled.

Personally, I think that the API is not equipped to give precise information on how trade Pokemon are received (i.e. which location/NPC/Pokemon to trade) so leaving the encounter method simply as `npc-trade` with no conditions isn't unreasonable. If we make changes to it in the future, that can be its own larger PR.

notblisy:
> I think adding an encounter condition for the Pokemon needed to trade is not a bad idea, but if a change like that is done for these encounters then a new PR should definitely be made for previous ones. While Duking is an easy enough character to recognize, not every trainer with trades in-game will have a name at all so I think a condition like that is not easily scaled.
> 
> Personally, I think that the API is not equipped to give precise information on how trade Pokemon are received (i.e. which location/NPC/Pokemon to trade) so leaving the encounter method simply as `npc-trade` with no conditions isn't unreasonable. If we make changes to it in the future, that can be its own larger PR.

A few things--

1. I think encounter conditions are the perfect use case for mentioning what Pokemon you need to bring with you, personally. It'd be trade-[X-Pokemon]. To me it's no different than the conditions for X-item or X-caught-articuno, who is right before my trades. I don't really think this is out of scope for encounter conditions. There's some pretty niche stuff in there. 

2. Locations for the trade is set up fine, it's a typical encounter. We would specify which location_area, like dukings house, to trade at. To me this is no different than how a gift Pokemon works, no? You go to Dukings house to get plusle in Colosseum, for example. There's just no condition of trading him a pokemon to do it.

3. I agree that all previous trades should have this condition met, however, there are basically no previous trade encounters documented. I made an issue here: https://github.com/PokeAPI/pokeapi/issues/1451

We would really only have to add conditions for the gen 7 trades, so to me this isn't as big of a deal as it seems to have to add these requirements.

I am amenable to removing the NPC name from the condition, Duking is one of very few named NPCs who you'd trade with. Jasmine is another in HGSS. 

If I did this it'd look like:

- 112,15,trade-wooper,0

Does that sound acceptable?

jemarq04:
> If I did this it'd look like:
> 
>     * 112,15,trade-wooper,0
> 
> 
> Does that sound acceptable?

I think this makes sense to me - @Naramsim what do you think?

Naramsim:
Hi, I think my previous comment stemmed from the facts that I do not yet fully grasp the difference between encounter methods and encounter conditions. It seemed to me that these aren't linked, but instead thanks to the _encounter_condition_value_map_ they are. I tried to understand a bit more the schema and now i'm putting together the pieces. 

Does it make sense to have the encounter condition listed as `npc-trade`, with the added `npc-` prefix? 

And in the `encounter_condition_prose.csv` I'd remove the exclamation mark from the added line. 

> * 112,15,trade-wooper,0
> Does that sound acceptable?

I think it is.

notblisy:
> o

Encounter conditions links to encounter value maps, and encounter value map links to encounters. 

They are basically little extra things you need to do to fulfill an encounter. Like a swarm or a slot 2 in the DS for a gengar in DPPt, etc. 

I'll make the change to remove dukings name. 

I don't think it makes sense to have it say NPC-Trade. it would then say NPC trade as location and then npc-trade-wooper as a condition. For me reading it as trade-wooper makes more sense when parsing the info. It's just the action that you take.

notblisy:
removed duking and hordels name from the encounter condition values. I also added two conditions to ho-oh in Colosseum. I didn't make new ones-- I just tagged it with entries 113 and 114 which I made for XD, but also happen to be conditions for Ho-Oh.

edit: also removed the !

jemarq04:
Alrighty, I'll start a full review soon

data/v2/csv/locations.csv
```diff
@@ -1087,3 +1087,12 @@ id,region_id,identifier
 1103,11,agate-village
 1104,11,the-under
 1105,11,mt-battle
+1106,11,pokemon-hq-lab
+1107,11,gateon-port
+1108,11,cipher-lab
+1109,11,cave-pokespot
```
jemarq04: For the implementation of the gen 6 mirage spots, we added a mirage spot location with each area as a related location area. Maybe we should similarly add a location `pokespot` with each of the variants as location areas? (e.g. `pokespot-cave` and `pokespot-oasis`.)

data/v2/csv/encounters.csv
```diff
@@ -65451,3 +65451,107 @@ id,version_id,location_area_id,encounter_slot_id,pokemon_id,min_level,max_level
...
+71107,20,1250,1372,213,20,20
+71108,20,1250,1372,246,20,20
+71109,20,1296,1368,152,5,5
+71110,20,1296,1368,155,5,5
+71111,20,1296,1368,158,5,5
+71112,20,1297,1368,175,25,25
+71113,20,1294,1367,299,26,26
+71114,20,1295,1367,299,26,26
+71115,20,1270,1367,299,26,26
+71116,20,1259,1367,299,26,26
```
jemarq04, line 65557: For these Nosepass encounters - are they meant to be snag-rematch or snag? They're listed as a snag encounter, but Bulbapedia often lists the rematches after the initial encounter so I wanted to check.

This is the only one listed in multiple locations in XD, which I thought was interesting.

data/v2/csv/encounters.csv
```diff
@@ -65451,3 +65451,107 @@ id,version_id,location_area_id,encounter_slot_id,pokemon_id,min_level,max_level
...
+71073,20,1288,1367,108,38,38
+71074,20,1288,1367,123,40,40
+71075,20,1288,1367,113,39,39
+71076,20,1289,1367,338,41,41
+71077,20,1289,1367,121,41,41
+71078,20,1289,1367,125,43,43
+71079,20,1289,1367,277,43,43
+71080,20,1289,1367,143,43,43
+71081,20,1289,1367,62,43,43
+71082,20,1289,1367,122,43,43
```
jemarq04:
```suggestion
71081,20,1289,1367,62,42,42
71082,20,1289,1367,122,42,42
```
Bulbapedia lists the encounters with Poliwrath and Mr. Mime at level 42

data/v2/csv/encounter_condition_values.csv
```diff
@@ -107,3 +107,9 @@ id,encounter_condition_id,identifier,is_default
 106,10,other-caught-articuno,0
 107,10,other-caught-zapdos,0
 108,10,other-caught-moltres,0
+109,15,trade-togepi,1
```
jemarq04: If I remember, can't you also trade a Togetic if you've evolved Togepi to get the Elekid?

jemarq04 (changes requested):
Got some time to revisit this - things look really good! Just minor comments this time. 

Also, I know this was from the last PR, but I notice that Bulbapedia calls it Phenac Stadium, where the location area uses `colosseum`. Could we rename that as well?

data/v2/csv/locations.csv
notblisy: Each pokespot is a distinct location. [Here's a pic of how it looks on the map](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9_Spot#/media/File:Orre_Rock_Pok%C3%A9_Spot_Map.png).

The reason I don't like it is because there's not one main area called "Pokespot" that these would be sub-areas of.

data/v2/csv/encounters.csv
notblisy: XD doesn't have rematches in the same way as Colosseum. In Colo you'd get rematches in the post-game only, pretty much. And they'd be in a unique, and different location, compared to where they were in the main story.

In XD rematches are done through Miror B. He is an NPC that roams around the map and will be in random locations with shadow rematches. 

Nosepass is in-fact the only Pokemon who is in a random location because his default spawn is with Miror B so he can be in a few spots.

data/v2/csv/encounter_condition_values.csv
notblisy: Oh, wow, I never even considered this. I think you're right. I can add a second condition for it. Alternatively, I can have it say "trade-togepi-or-togetic".

Thoughts?

data/v2/csv/encounters.csv
notblisy: Serebiis tables have betrayed me again. I thought I caught all the errors. Thanks!

data/v2/csv/locations.csv
jemarq04: The reason I suggest this is because we've done the same in the past with `mirage-spot` as the location and the different variants of it as the location areas. (I had also assumed that you were planning something similar for the hidden grottoes you mentioned in another issue recently?)

I do see your point for how they are distinct locations, but I think this would be more in-line with what we've done in the past. Even Bulbapedia has them as sub-sections to the Pokespot page for example.

data/v2/csv/encounters.csv
jemarq04: Ah ok, gotcha!

data/v2/csv/encounter_condition_values.csv
jemarq04: I think `trade-togepi-or-togetic` should work fine!

data/v2/csv/locations.csv
notblisy: For the hidden grottos I was planning on making them each location areas of the route they're on, but we can discuss that on that issues page.

I can make this change, I don't feel too strongly about it.

jemarq04 (approved):
Looks great! Thanks again!

notblisy:
Thanks for checking me as usual.

Naramsim:
Thank you both!

pokeapi-machine-user:

pokeapi-machine-user:

data/v2/csv/location_area_encounter_rates.csv
```diff
@@ -3877,3 +3877,12 @@ location_area_id,encounter_method_id,version_id,rate
 1206,1,7,10
 1206,1,8,10
 1206,1,9,10
+1270,53,20,50
+1270,53,20,35
+1270,53,20,15
```
Dropries, line 3882: Shouldn't the `rate` be unique for each `location_area_id` * `encounter_method_id` * `version_id`? 
Otherwise, it seems to be a duplicate of what you added in `data/v2/csv/encounter_slots.csv`.

According to the [documentation](https://pokeapi.co/docs/v2#encounterversiondetails), this is supposed to be "the chance of a pokespot encounter to occur in xd".

Image

notblisy: I think you're right. A Pokespot rate should be 90% because of the 10% munchlax chance. I was confused as to the difference here.

jemarq04: > Shouldn't the `rate` be unique for each `location_area_id` * `encounter_method_id` * `version_id`? Otherwise, it seems to be a duplicate of what you added in `data/v2/csv/encounter_slots.csv`.
> 
> According to the [documentation](https://pokeapi.co/docs/v2#encounterversiondetails), this is supposed to be "the chance of a pokespot encounter to occur in xd".

Good catch!

notblisy: https://github.com/PokeAPI/pokeapi/pull/1468 

Pushed a PR to fix this

HybridShivam:
Thanks for adding the XD data but wouldn't it be nice to have the received data actually sorted by the version-id?
Now "XD" is at the bottom of the list, this is not ideal. I have also seen this kind of sequence breaking in `pokemon-species/flavour_text_entries`, so I raised an issue #1488. 


@Naramsim Maybe this is a bigger issue with how API handles new data.

Beedrill
Screenshot (516)

Naramsim:
Can you try for Beedrill only to move the resources up in the CSV file and see if the API as well moves the property up? 

I know that python doesn't guarantee the ordering of lists so that's as well a point to keep in mind

PR #1466, opened 2026-04-02
item mechanics initial setup

ninjaman0722:
THIS IS INCOMPLETE AND IS ONLY A REFERENCE FOR ISSUE #1460!


jemarq04:
You can mark the PR as a draft so that we know it's not ready for review

ninjaman0722:
> You can mark the PR as a draft so that we know it's not ready for review

I will as soon as I get home!

Although its not ready to be used yet, I would like someone to look over it and make sure that what I have so far is good. Its easy to change stuff now before I get too far into this lol

When I get home I will also come up with a detailed explanation as to how this works

PR #1499, opened 2026-04-19
Add Gen IV Great Marsh changing encounters

### LINKED ISSUE #1496: Great marsh in D/P/Pt missing encounters for the changing pokemon
The Great Marsh in Gen IV has [changing encounters](https://bulbapedia.bulbagarden.net/wiki/Great_Marsh#Changing_Pok%C3%A9mon_2) in the 6 areas.

I'm not sure what the best way to model this is. My first idea was to create a 10% chance encounter slot for D/P and Platinum (separate) and have encounter condition values for each of the pokemon, but that doesn't seem right. Plus, we need more condition values because of pre and post national dex.

Can we have a bunch of pokemon in the same encounter slot with no different encounter condition values? Would that just mean its only 'one' of these pokemon in that slot that have the 10% chance? That could work.

Machine-Maker:
# Change description

Closes #1496 
- Adds new condition type for great marsh selection, x of 32 options.
- Adds encounters for all the changing pokemon, taking into account pre and post national dex states.

Claude Code helped me analyze decompiled [diamond](https://github.com/pret/pokediamond) and [platinum](https://github.com/pret/pokeplatinum) repos. Also generated the big huge changes based on how they work.

## Contributor check list


data/v2/csv/encounter_condition_values.csv
```diff
@@ -177,3 +177,11 @@ id,encounter_condition_id,identifier,is_default
 176,17,friend-safari-slot-1,1
 177,17,friend-safari-slot-2,1
 178,17,friend-safari-slot-3,1
+179,18,great-marsh-daily-slot-0-of-32,1
```
jemarq04: I'm not sure I like how these are named (but I'm not sure how else to describe the difference in rates) so we can leave that. That being said, what would a rate of 0/32 for a daily spawn mean? I think this was likely a mistake.

jemarq04: Ah, if this is for the default field I'd rather something similar to the `story-progress` condition and name it `great-marsh-daily-slot-none`, as if you don't check the binoculars you don't have any daily spawn added in to the encounter pool.
```suggestion
179,18,great-marsh-daily-slot-none,1
```

data/v2/csv/encounter_condition_value_prose.csv
```diff
@@ -298,4 +298,12 @@ encounter_condition_value_id,local_language_id,name
...
\ No newline at end of file
+178,9,Randomly assigned to Friend Safari Slot 3
+179,9,0 of 32
+180,9,1 of 32
+181,9,2 of 32
+182,9,3 of 32
+183,9,5 of 32
+184,9,Before National Pokédex
+185,9,4 of 32
+186,9,15 of 32
```
jemarq04: I think these should be a bit more verbose, like the other entries in the file

jemarq04 (changes requested):
I've been taking a look at these but I think before I do an in-depth review I'd like you to manually go through these yourself. From an initial glance, rates seem to be incorrect and I'm already seeing duplicates from my first look at Area 1 (e.g. Wooper).

I've made a couple other comments below.

Machine-Maker:
It might look like there are duplicates, but that's just cause there are 2 encounter slots for each area. So we need them to be possible on each slot.

Say for diamond, for example. Each of the "special" pokemon, when selected, have a 10% encounter rate in that area. It's the same for every pokemon. But that 10% is distributed across 2 different encounter slots. This is because it *decreases* the encounter rate for 2 separate pokemon by 5% each (by replacing that slot). So there are a lot of what seem like duplicates, but they are just in different encounter slots.

The same is true for platinum, there are 2 5% "wooper" encounters that are replaced by the 2 5% encounters for the selected pokemon.

For D/P, its either encounter_slot_id 7 or 8

For Plat, its either encounter_slot_id 129 or 130.

data/v2/csv/encounter_condition_values.csv
Machine-Maker: I'm fine with either really, but I just feel like `none` doesn't really convey that this is impossible. Like this condition is never true. It's my understanding that there is *always* a selected pokemon in each area and so the 2 5% encounter slots for the default pokemon are just never ever true. So I don't know if `none` is the best way to convey that (maybe 0-of-32 isn't the best either).

jemarq04: If the daily spawn is always set, a slot of `none` would definitely convey it's not possible. `0-of-32` is a convention used only for these IDs, so that's not immediately conveying that this would be impossible either.

Plus, I believe if you don't check the binoculars, the daily spawn isn't set. So this condition would be possible.

jemarq04:
> It might look like there are duplicates, but that's just cause there are 2 encounter slots for each area. So we need them to be possible on each slot.
> 
> Say for diamond, for example. Each of the "special" pokemon, when selected, have a 10% encounter rate in that area. It's the same for every pokemon. But that 10% is distributed across 2 different encounter slots. This is because it _decreases_ the encounter rate for 2 separate pokemon by 5% each (by replacing that slot). So there are a lot of what seem like duplicates, but they are just in different encounter slots.
> 
> The same is true for platinum, there are 2 5% "wooper" encounters that are replaced by the 2 5% encounters for the selected pokemon.
> 
> For D/P, its either encounter_slot_id 7 or 8
> 
> For Plat, its either encounter_slot_id 129 or 130.

Ok, thanks for the explanation. I've done a more thorough review and the encounters are looking good - thank you for adding these in! All that's left is to address the previous comments I left. I'll take a look at your other PRs soon as well.

data/v2/csv/encounter_conditions.csv
```diff
@@ -15,4 +15,5 @@ id,identifier
 14,special-encounter
 15,trade
 16,coins
-17,friend-safari-slot
\ No newline at end of file
+17,friend-safari-slot
+18,great-marsh-daily-slot
```
jemarq04, line 19: I also missed this in my earlier review - please add the corresponding entry in `encounter_condition_prose.csv`

Machine-Maker: I think I already did this? 
https://github.com/PokeAPI/pokeapi/pull/1499/changes#diff-f8931629eb4d9b34d999860ca59f189a72d128a9ba16da4a2c5f50ccb7a4ad06R39

jemarq04: Ugh, yes you did, I need to double-check my double-checks. Thanks!

data/v2/csv/encounters.csv
```diff
@@ -66084,6 +66084,7 @@ id,version_id,location_area_id,encounter_slot_id,pokemon_id,min_level,max_level
 71742,12,23,1396,349,10,20
 71743,13,23,1396,349,10,20
 71744,14,23,1397,349,10,20
+
```
jemarq04:
```suggestion
```
Looks like an extra line made it through!

data/v2/csv/encounter_condition_value_prose.csv
Machine-Maker: Done

data/v2/csv/encounter_condition_values.csv
Machine-Maker: Done

jemarq04 (changes requested):

data/v2/csv/encounters.csv
Machine-Maker: done

jemarq04 (approved):

Naramsim:
Oh no! :((( Merge conflicts :)

sorry guys!

Machine-Maker:
Ok, fixed conflicts, shifted all the ids down.

jemarq04 (approved):

pokeapi-machine-user:

pokeapi-machine-user:

PR #1525, opened 2026-05-19
Add missing evolution entries

FreakMediaLP:
# Change description

Added missing evolution locations:
- "New Mauville" for "Magneton" and "Probopass" (available in ORAS)
- "Petalburg Woods" for "Leafeon" (available in ORAS)
- "Shoal Cave" for "Glaceon" (available in ORAS)

## AI coding assistance disclosure

Used AI to get a better understanding of the `pokemon_evolutions.csv` and its connections with other .csv files.
All changes were made manually though.

## Contributor check list

- [x] I have written a description of the contribution and explained its motivation.
- [x] I have written tests for my code changes (if applicable).
- [x] I have read and understood the [AI Assisted Contribution guidelines](https://github.com/PokeAPI/pokeapi/blob/master/CONTRIBUTING.md#ai-assisted-coding).
- [x] I will own this change in production, and I am prepared to fix any bugs caused by my code change.

PR #1547, opened 2026-06-07
Add evolution condition: near_special_rock

jemarq04:
# Change description

Follow-up to PR #1525 and discussion #1538. Adds an additional evolution condition `near_special_rock` to `pokemon_evolution.csv` for early Leafeon and Glaceon evolutions near Moss Rock and Icy Rock, respectively. Set to `0` for all other evolutions.

## Contributor check list


Naramsim:
I'm not super convinced by the line 

            near_special_rock=bool(int(info[21])),

What do you think about adding a new different catch-all attribute such as 'special condition' that we can reuse also for other evolutions? This would be more complicated of course and would need some prose as well. 

Just asking

jemarq04:
> What do you think about adding a new different catch-all attribute such as 'special condition' that we can reuse also for other evolutions? This would be more complicated of course and would need some prose as well.

Something like that would be nice, but we have no way of adding any prose or additional comments on these evolution condition fields. That's why in the past we've just added relevant new fields (e.g. `needs_multiplayer`, `needs_overworld_rain`, and `turn_upside_down`). As far as I know, the framework doesn't have any way to add prose on these evolution conditions. 

Talking about this did remind me of the `evolution_triggers.csv` file – maybe we could add a `level-up-near-moss-rock` and `level-up-near-icy-rock` triggers for these evolutions instead? This would allow for dedicated prose that we could write.

Naramsim:
Or we could modify the schema to support the prose for a new "evolution condition" field. 

Are there many "condition s" to evolve a Pokemon? That we currently don't support? Can you think of any?

jemarq04:
No, I don't believe we have any more that aren't accounted for. I made a PR some time ago that added some evolution fields and triggers to account for any placeholder uses of the `other` evolution trigger, so everything should be handled aside from these special Leafeon/Glaceon evolutions.

Naramsim:
Oh well then let's go with this PR 🙂 thanks a lot!

Naramsim (approved):

Naramsim:
If you have time could you also update the docs? 

I'll update the gql

pokeapi-machine-user:

pokeapi-machine-user:

jemarq04:
> If you have time could you also update the docs?
> 
> I'll update the gql

Of course, I’ll update the docs! I hadn’t realized there’s another step that needs to be taken with these changes - is this a makefile target or something I should be running when I make changes to CSV structure?

PR #1553, opened 2026-06-10
feat: Migrate package manager to uv

FallenDeity:
# Change description

First set of pr for migrating and adopting modern python tooling ecosystem, this pr attempts to migrate from pip to uv for package management as discussed in #1391 & #1521 it breaks down the changes @phalt initially made to smaller more iterative parts

cc: @phalt @Naramsim 

Used a bit for reviewing the changes and coverage if I missed any transitive command update

## Contributor check list

- [x] I have read and understood the [AI Assisted Contribution guidelines](hxtps://github.com/PokeAPI/pokeapi/blob/master/CONTRIBUTING.md#ai-assisted-coding).


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 jobs:
   test:
     docker:
-      - image: cimg/python:3.13.7
```
Naramsim, line 6: Can we use a matrix here and test LTS and the newest python version?

.python-version
```diff
@@ -0,0 +1 @@
+3.14
```
Naramsim, line 1: Do we need this file? I'd like to support multiple Python versions.

docker-compose.yml
```diff
@@ -60,7 +54,7 @@ services:
     restart: always
     environment:
       HASURA_GRAPHQL_DATABASE_URL: postgres://${POSTGRES_USER:-ash}:${POSTGRES_PASSWORD:-pokemon}@db:5432/${POSTGRES_DB:-pokeapi}
-      HASURA_GRAPHQL_ENABLE_CONSOLE: "true"
+      HASURA_GRAPHQL_ENABLE_CONSOLE: "false"
```
Naramsim, line 57: Why false?

.python-version
FallenDeity: well there are a few benefits to having the interpreter version pinned, the pipeline uv sets up automatically + make install/setup (uv sync) auto installs the suitable python version

I think having a pinned python version for the project is fine since uv handles that anyways and to run it with diff versions locally the developer can either use the `--python` flag with uv or modify the version pin temporarily https://docs.astral.sh/uv/concepts/python-versions/#requesting-a-version

instead in the pipeline we can have the matrix which runs the project against diff version and the dev can later modify it locally we can make a note in the readme

docker-compose.yml
FallenDeity: I moved all the dev settings to the -dev compose file and left this as barebones base compose file, the graphql console is not really a recommended setting to be enabled to be in prod/ is a local/dev based setting so figured overriding it in the dev compose file is a more suitable location for it

is there any particular reason for us to have it globally set to true I am not that familiar with graphql but shouldnt the console be enabled in a much more limited manner or does it not matter in our case?

.circleci/config.yml
FallenDeity: added a matrix in circle ci for `make test` and one in database.yml `sqlite` job for `make build-db` which are python version dependent moved the formatting and openapi spec generation to a separate step in circleci since they are not python version dependent

.github/workflows/database.yml
jemarq04: I'm not as familiar with the docker and `pg_dump` commands in this YAML, but this seems separate from the UV shift. I'm not opposed, but could you describe what the changes here would do?

Resources/docker/app/Dockerfile
```diff
@@ -1,35 +1,37 @@
-FROM python:3.13.7-alpine AS builder
+FROM python:3.14-alpine as builder
```
jemarq04: Should we be concerned with the automated warnings from Github on this line?

.python-version
jemarq04: I think adding a default python version is probably a good idea, but I agree that we should test older versions. I think adding them to the testing matrix is a good approach.

Makefile
jemarq04, line 41: Less a question directed at you, but is there a reason we have two targets that do the same thing (setup and migrate)?

.github/workflows/database.yml
FallenDeity: It was originally the same changes as here https://github.com/PokeAPI/pokeapi/pull/1391 in the og pr from paul but I found a few discrepancies in that pr such as it spins up a github actions services postgres, a redundant sqlite db from normal host run and a docker postgres build so I modified the current one to just do everything in the docker container itself (I didnt encounter any memory limit issues)

difference from the actual current running flow is that I removed the container teardown which happens rn and instead I just docker exec the drop and recreate for db which should speed up workflow times and avoid all the dump file copying between container and host /tmp

Resources/docker/app/Dockerfile
FallenDeity: it was a stylistic warning to use `AS` instead of `as` resolved this

Makefile
FallenDeity: I think it was orignaly created to apply schema changes to the db? based of on what I see in the readme https://github.com/search?q=repo%3APokeAPI%2Fpokeapi%20make%20migrate&type=code in the db migration section

and coincidentally make setup seems to does the same thing so perhaps it was left for the sake of being verbose or clarity of operation

.python-version
FallenDeity: image

yep added test matrixes

.github/workflows/database.yml
jemarq04: Ahh ok I see, thanks!

jemarq04 (approved):

.python-version
phalt: +1 for keeping the `.python-version` file. It is a common convention and helps ensure local dev works as expected.

.github/workflows/database.yml
phalt: Just so you know, I am also very unfamiliar with docker so a lot of my prior art is me wrangling things trying to get it to work (and before I was comfortable using llms so I was just hacking away). Don't take what I wrote as expert-level approaches 😆

Naramsim:
I want to review this PR, don't merge please. I'm slow I know, bear with me

Makefile
```diff
@@ -4,51 +4,55 @@ docker_config = --settings=config.docker-compose
...
 .SILENT:
 
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?# .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-install:  # Install base requirements to run project
-	$(PIP) install -r requirements.txt
+check-uv:
+	@command -v uv >/dev/null 2>&1 \
+		&& echo "uv is installed: $$(uv --version)" \
```
Naramsim: If `uv` is installed there's no need to print the version. This is just a depencency target, it's not gonna get executed directly.

pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "pokeapi"
+description = "The Pokemon API"
+version = "0.1.0"
+requires-python = ">=3.10,<3.15"
+readme = "README.md"
+license = { file = "LICENSE.md" }
+dependencies = [
+    "coverage==7.13.1",
```
Naramsim: `coverage` as well then can be a dev dependency.

.github/workflows/database.yml
```diff
@@ -21,52 +21,71 @@ jobs:
...
   postgres:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           submodules: recursive
-      - name: Build
+      - name: Build and start services
```
Naramsim, line 69: This is not building anymore, just starting the services.

.github/workflows/database.yml
```diff
@@ -21,52 +21,71 @@ jobs:
...
-      - name: Import database
+        run: docker compose exec -T db pg_dump -U ash -Fc -N 'hdb_*' pokeapi > pokeapi.dump
+      - name: Drop and recreate database
         run: |
-          docker compose cp ./pokeapi.dump db:/tmp/
-          docker compose exec -T -u postgres db sh -c "cd /tmp && pg_restore -h localhost -U ash -d pokeapi pokeapi.dump"
+          docker compose exec -T db psql -U ash -d postgres -c "DROP DATABASE pokeapi WITH (FORCE);"
+          docker compose exec -T db psql -U ash -d postgres -c "CREATE DATABASE pokeapi;"
+      - name: Import database
+        run: docker compose exec -T db pg_restore -U ash -d pokeapi < pokeapi.dump
```
Naramsim, line 83: Can we put back `-h localhost`? I like the commands to be descriptive, so that anyone can see the pipeline and use it as a piece of documentation.

.github/workflows/database.yml
```diff
@@ -21,52 +21,71 @@ jobs:
...
       - name: Test data
-        run: curl -Ss http://localhost/api/v2/pokemon/1/ | grep -q 'bulbasaur'
+        run: |
+          name=$(curl -fsSL http://localhost/api/v2/pokemon/1/ | jq -r '.name')
```
Naramsim: Also here I'd put the test as it was before

.github/workflows/docker-k8s.yml
```diff
@@ -61,6 +63,8 @@ jobs:
           platforms: local
           tags: pokeapi/pokeapi:local
           labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha,scope=docker
```
Naramsim: Can we remove all this caching and open later a separate PR? So It's gonna be easier to see how well/fast the caching works.

Resources/docker/app/Dockerfile
```diff
@@ -1,35 +1,37 @@
...
 
-RUN mkdir /code
 WORKDIR /code
 
-ADD requirements.txt /code/
-RUN apk add --no-cache postgresql-libs libstdc++
-RUN apk add --no-cache --virtual .build-deps gcc g++ musl-dev \
-    postgresql-dev binutils rust cargo && \
-    python3 -m pip install -r requirements.txt --no-cache-dir
+RUN --mount=type=cache,target=/root/.cache/uv \
```
Naramsim, line 11: same here, let's remove the cache mount and add it later in a separate PR.

docker-compose.yml
```diff
@@ -15,7 +14,7 @@ services:
     volumes:
       - pg_data:/var/lib/postgresql/data
     healthcheck:
-      test: pg_isready
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-ash} -d ${POSTGRES_DB:-pokeapi}"]
```
Naramsim: Does the healthcheck run even without `CMD-SHELL`? With only `pg_isready -U $${POSTGRES_USER} -d $$POSTGRES_DB}`

pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "pokeapi"
+description = "The Pokemon API"
+version = "0.1.0"
```
Naramsim: Can we add the current version? 2.9.0?

Or maybe bump to 2.10/3.x?

Naramsim (changes requested):

.github/workflows/database.yml
FallenDeity: the default behavior for dump and restore for postgresql isnt localhost or the local loopback address instead it uses the unix socket (file based) which is faster without the tcp/ip overhead I can force it to use localhost if required though but imo there is no particular reason to go through localhost

previously the docker invocation was using `-u postgres` (os system user) and the the command was using `-u ash` which likely touched upon this auth permission rule with unix socket file perms vs tcp/ip bypass https://www.postgresql.org/docs/current/auth-trust.html due to differing users in the command and db role access

> Local TCP/IP connections are not restricted by file-system permissions

```
docker compose exec -T -u postgres db sh -c "cd /tmp && pg_restore -h localhost -U ash -d pokeapi pokeapi.dump"
```

in the  new workflow the commands run as the container default execution user so we dont run into any permission misalignments for file r/w

Resources/docker/app/Dockerfile
FallenDeity: lets leave this cache here? since its a uv or a requirements dependency cache ensuring uv dosent have to reach out to pypi and redownload all packages if there is a change in pyproject or uv lock its recommended in their docs too https://docs.astral.sh/uv/guides/integration/docker/#caching

Ill remove the ci/cd or workflow layer image caching mentioned above though for a future pr

docker-compose.yml
FallenDeity: well currently the compose expansion will be taking place where it takes env vars from the compose file and replaces them before container starts so `cmd-shell` is not required since no shell expansion takes place just `cmd` too would work. 

As you mentioned this works too `pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}` because docker in a string format command spins up an implicit shell and will perform a shell expansion during runtime

.github/workflows/database.yml
FallenDeity: docker compose up does build the image too if it dosent exist (which it dosent when this task first runs) but ill also add a `--build` flag to be explicit

Makefile
FallenDeity: removed the echo

pyproject.toml
FallenDeity: moved to dev group

.github/workflows/database.yml
FallenDeity: reverted

.github/workflows/docker-k8s.yml
FallenDeity: removed workflow caching steps

pyproject.toml
FallenDeity: bumped to `2.10.0` since its not a breaking changes just an infra change, apis remain same

FallenDeity:
was on vacation last week 🏖️ so changes/followup were bit delayed

.github/workflows/database.yml
Naramsim: ok, let's keep it as it is then :) Thanks! I didn't know about this sockets vs tcp

Naramsim (approved):

Naramsim:
Great! @FallenDeity your contributions are helping a lot! Thanks also to @phalt that submitted part of the work in another PR.

pokeapi-machine-user:

pokeapi-machine-user:

PR #1557, opened 2026-06-13
Add Johto safari zone encounters

### LINKED ISSUE #1522: Johto Safari Zone is missing block-based encounters
Johto Safari zone is missing its block-specific encounters. These are encounters that replace base encounters if certain conditions are met. 

I'll fix this, initially by just having a condition type with 2 values, for "selected" and "not selected". We can add like 60 condition values later if we really want to for the correct point requirements.

Machine-Maker:
# Change description

Adds gen iv Johto safari zone encounters
Closes https://github.com/PokeAPI/pokeapi/issues/1522

Claude helped me understand the [decompiled codebase](https://github.com/pret/pokeheartgold) for heartgold. And to design a script that would generate all the encounter lines.

## Contributor check list


jemarq04:
Can you provide more of a description to what you're adding? What I'm reading on Bulbapedia does not match what you're adding here.

For example, Fearow isn't available in `johto-safari-zone-peak` unless there are 5 Forest block objects placed in it, according to [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Johto_Safari_Zone#Pok%C3%A9mon). However, this PR just adds a `johto-safari-displaced` with prose "Displaced by block-based encounter". This doesn't seem very clear to me (and I'm not sure what is even meant by "displaced").

Some encounters require multiple types of blocks, so maybe something more reasonable would be a set of encounter conditions - one for each type of block object that would need to be placed. Then a set of relevant condition values would be added (e.g. a `johto-safari-forest-block` condition with a `johto-safari-forest-block-5` condition value for Fearow's, or something like this). I think this would make things a lot clearer as to how we would encounter these pokemon in the area

Also.. "Claude helped" is not much of a description for how you used AI to aid this PR. You've added a lot of helpful encounters we've missed, so thank you for that, but please try to be descriptive in these descriptions so we know what we're reviewing

Machine-Maker:
I adjusted the ai disclosure statement.

As for the encounter structure, I originally tried to model the block count requirement for each pokemon, but that quickly got out of hand, especially with so many different combinations. Some pokemon require amounts of 2 different blocks as well which only further compounds the issue.

So I decided, at least for now, to go back and simplify it. I would just have one new encounter condition type with 2 values, "displaced" or "not displaced". You need some condition on the base encounters for each safari zone biome that says this encounter can be turned off, or "displaced" by a block-based encounter. You don't technically need this for *all* the base encounters for each area, but there's not a good way of knowing which you do, and which you don't. The base encounters are replaced by block-based encounters just in the order the base encounters are stored in, one-by-one as new block-based encounters are activated. 

As usual it seems, my naming can be a bit off and I suppose while I was working on it, "displaced" just seemed to be natural to me, but of course I am open to alternatives. I've made some changes already, and can make more.

It basically comes down to us needing 2 condition values, one to say that this encounter (a base encounter in an area) *can* be turned off if its replaced by one of the block-based encounters. And we need another condition value that says this encounter is only activated by placing various blocks (without having to model the exact block requirements).

jemarq04:
> As for the encounter structure, I originally tried to model the block count requirement for each pokemon, but that quickly got out of hand, especially with so many different combinations. Some pokemon require amounts of 2 different blocks as well which only further compounds the issue.
> ...
> It basically comes down to us needing 2 condition values, one to say that this encounter (a base encounter in an area) _can_ be turned off if its replaced by one of the block-based encounters. And we need another condition value that says this encounter is only activated by placing various blocks (without having to model the exact block requirements).

My only concern with this condition is that it amounts to saying "this encounter can happen if it's possible", which seems unhelpful. I understand that it would be complicated, but unfortunately the encounter conditions in-game are just complicated. To describe something complex, we'd need lots of encounter conditions to account for it. I think what I proposed earlier would involve a lot of encounter condition values, but it would be the most accurate representation of the encounters in the games. To clarify, this isn't as much about the word `displaced` or `blocks-required` as it is about the clarity in our data.

Machine-Maker:
Ok, re worked to have more specific conditions.

data/v2/csv/encounters.csv
```diff
@@ -71067,3 +71067,599 @@ id,version_id,location_area_id,encounter_slot_id,pokemon_id,min_level,max_level
...
+77113,16,813,654,436,46,46
+77114,15,813,655,80,45,45
+77115,16,813,655,80,45,45
+77116,15,813,656,363,45,45
+77117,16,813,656,363,45,45
+77118,15,813,656,363,45,45
+77119,16,813,656,363,45,45
+77120,15,813,656,363,44,44
+77121,16,813,656,363,44,44
+77122,15,813,657,119,20,20
```
jemarq04: Reviewing the peak area everything seemed good, but where did you get the encounters using old/good/super rod and surfing? None of these are listed on Bulbapedia. There's a surfing encounter for Surskit that also doesn't seem possible in any of the areas, so I don't think there was just a location area mixup.

data/v2/csv/encounter_conditions.csv
```diff
@@ -24,3 +24,4 @@ id,identifier
 23,save-data
 24,alolan-diglett-found
 25,weather
+26,johto-safari-block-replacement
```
jemarq04: This is a much cleaner idea than what I suggested – I don't know why I hadn't thought of making *all* of them condition values under one condition. Good call!

That said, for consistency we like to have the encounter condition value identifiers start with the encounter condition it comes from. Would you be able to rename your condition/condition values to match that? Something like `johto-safari-blocks` and `johto-safari-blocks-min-49-water` or something like that could work, but I'll leave it up to you to find something you like

jemarq04 (changes requested):

data/v2/csv/encounters.csv
Machine-Maker: Yeah, I was using the decompile source [here](https://github.com/pret/pokeheartgold/blob/master/files/arc/safari_enc.json#L3498), that actually has water-based encounters listed for the peak area... but you are right, I don't think there's actually any water to use these. 

The script I wrote just assumed all the encounters in the decompile were valid, but I'll filter them out based on whatever areas have wate.r

data/v2/csv/encounter_conditions.csv
Machine-Maker: done

jemarq04 (approved):
Looks great - thank you so much for adding these encounters!

When reviewing the encounters, I did notice that the location areas are named a bit silly so we get things like `johto-safari-zone-zone-desert` with 'zone' repeated. If you'd be willing, could you rename these location areas (813-824) to either remove 'zone' or move it to the end? Up to you how to do it!

Naramsim:
Thank you both!

Issue #1569, opened 2026-06-24
"order" of Pokémon forms is shared between unrelated species


SnorlaxMonster:
At present, the `order` column of the `pokemon_forms.csv` is mostly just an incrementing list that groups all forms of the same species together. This makes it very inconvenient to insert new forms for pre-existing Pokémon, as every subsequent entry needs to have its `order` incremented.

However, while the list at first appears to be an incrementing list (with a unique value per form), this is not actually the case (as I discovered in https://github.com/PokeAPI/pokeapi/pull/1532). [The documentation](https://pokeapi.co/docs/v2#pokemon-forms) does note
> Multiple forms may have equal order, in which case they should fall back on sorting by name.

The first major group of non-unique entries is Alcremie. The Alcremie forms have distinct `order` values for each Cream, but each Cream shares the same `order` value across Sweets. So `alcremie-vanilla-cream-berry-sweet` is 1256 and `alcremie-ruby-cream-berry-sweet` is 1257, but `alcremie-vanilla-cream-love-sweet` is also 1256.

However, after Alcremie, there are a bunch of DLC Sword & Shield Pokémon, Hisui Pokémon, and Paldean Pokémon that share `order` values with Gigantamax Pokémon and forms introduced in Hisui and Paldea. Some examples:

order | Entry 1 | Entry 2
-- | -- | --
1292 | eternatus-eternamax | venusaur-gmax
1293 | kubfu | charizard-gmax
1294 | urshifu-single-strike | blastoise-gmax
1313 | fuecoco | flapple-gmax
1314 | crocalor | appletun-gmax
1327 | rabsca | growlithe-hisui
1328 | greavard | arcanine-hisui
1373 | naclstack | miraidon-aquatic-mode
1374 | garganacl | miraidon-glide-mode

## Proposal

My proposal is that the order column do something similar to what Pokémon Champions does internally: Concatenate a National Pokédex number and a form ID to produce a unique identifier for a specific form of a Pokémon. For example, Venusaur (Pokédex #0003) would be form ID 00003000, while Mega Venusaur would be form ID 00003001, and Charmander (Pokédex #0004) would be form ID 0004000. (Since these are integers, the leading 0's would be stripped, but I've included them here to illustrate how the numeric identifiers are based on Pokédex numbers.)

This could even replace the existing `id` column of the `pokemon_forms.csv` table, and the `order` column could just be deprecated.

jemarq04:
[...]

PR #1571, opened 2026-06-25
Construct "order" for forms from Pokédex No. & form index

SnorlaxMonster:
# Change description

Addresses https://github.com/PokeAPI/pokeapi/issues/1569. This constructs the `order` column of `pokemon_forms.csv` by concatenating a three-digit form number with the Pokédex number. This should facilitate inserting new forms in the future, without needing to reorder every subsequent Pokémon.

Note that because I wanted the base forms to be form ID 0, I have also shifted all of the `form_order` values by -1. Additionally, some of the `form_order` values also weren't unique previously, so those have been modified to be unique.

No AI used.

## Contributor check list

- [N/A] I have written tests for my code changes (if applicable).


jemarq04:
This is great! Did you do this with a script? If so, it would be great to have this in the API when new forms are added. You can add it into `Resources/scripts/data`.

data/v2/csv/pokemon_forms.csv
```diff
@@ -1,1580 +1,1580 @@
...
-10546,glimmora-mega,mega,10321,31,1,1,1,2,1377
-10547,tatsugiri-curly-mega,mega,10322,31,1,1,1,4,1359
-10548,tatsugiri-droopy-mega,mega,10323,31,1,1,1,5,1360
-10549,tatsugiri-stretchy-mega,mega,10324,31,1,1,1,6,1362
-10550,baxcalibur-mega,mega,10325,31,1,1,1,2,1357
-10551,pyroar-female,female,668,15,0,0,0,2,949
-10552,frillish-female,female,592,11,0,0,0,2,794
-10553,jellicent-female,female,593,11,0,0,0,2,795
-10554,meowstic-female-mega,mega-female,10326,31,1,1,1,4,984
+1,bulbasaur,,1,28,1,0,0,0,1000
```
Naramsim, line 2: Hi, I saw your comment for the `start-at-0` but I think we should leave the order starting from 0. In all the other CSVs we have it so. 

I also don't like the order starting at 1000, it's seems like a sort of a quick hack. If this is done by a script maybe we could simply keep the order as it is now and run the script each time we update this file, what do you think about it?

Secondly, if this is done by a script we could even integrate it in the build logic and have the order column created at build-time. This is of course more difficult to implement and needs some code knowledge. 

I'm not entirely opposed to the `1000` id system I'd just want to express my ideas. 

My bad for not having participated to the https://github.com/PokeAPI/pokeapi/issues/1569 issue.

jemarq04: This was originally brought up in a different PR (#1532) by the author: 
> (IMO, this would be better handled similarly to what Champions does internally, where they just allocate the last 3 digits to the form ID and the first 4 to the National Pokédex number, so that inserting a new form doesn't require incrementing all of the others; e.g. Mega Venusaur is 0006001. However, changing that structure would belong in a separate PR, so I've just stuck to what is already there.)

Of course, I think we can make use of some nice scripts to do it for us in any format we'd like, so it really goes to preference.

This is something that Champions implemented, and personally I think it's more readable and easier to maintain manually. A script could be used to help maintain it, but if a new Mega Bibarel were introduced tomorrow we could simply go in and say "easy, that's order 400001 for national pokedex number 400 and form 001" without having to run a script.

I do like the idea of doing it at build-time as well, avoiding the need to write them manually into the CSV file. Whichever format we choose this would be a nice next step.

SnorlaxMonster: Having the `order` be generated at build time, rather than specified in the input CSV files, sounds like a good idea. Although it's a much more involved change than in this PR.

> Hi, I saw your comment for the start-at-0 but I think we should leave the order starting from 0. In all the other CSVs we have it so.

Are you trying to say that you would prefer that `form_order` starts from 1 instead of 0? I had deliberately changed it to be 0-indexed so that the default forms were all form 0 (and thus Bulbasaur is `0001000`, etc.), which aligns with the internal numbering of Champions as well. But I can change it back to being 1-indexed if you prefer. (Especially if `order` is just generated at build time instead, it doesn't really make much difference whether this field is 0-indexed or 1-index.)

Naramsim: Yes, I don't think we need to support pokemon champions ordering system. I'd prefer to keep consistency across the API. So let's revert back to 1-index. 

Can you investigate a bit if it's feasible building the order at runtime? If it's relatively easy we can go that way. If not, I would like the order to be subsequent as it is now. I don't know if your script is easy to modify in this regard. Last option is to have big spaced numbers.

Naramsim: Hey @SnorlaxMonster are you still there? Sorry for the ping

PR #1572, opened 2026-06-25
add latest berries and alter schema to allow empty properties from th…

programgames:
#1126 
I added both hopo berry and roseli item.

Also Kee,Marangua,Hopo and Roseli berries in the berry table.
As you said in the issue, new berries don't have much properties as old ones do I had to change the schema to allow empty values.

Datas came from Serebii and bulbapedia :
- https://www.serebii.net/itemdex/keeberry.shtml
- https://www.serebii.net/itemdex/marangaberry.shtml

and then : 
- https://www.serebii.net/itemdex/hopoberry.shtml
- https://www.serebii.net/itemdex/roseliberry.shtml

I would like to know if it a good practice to add new data add the end of the file like I did.

Thank you

PR #1621, opened 2026-07-24
feat: Add new model and data for Encounter Pokemon Details

jemarq04:
# Change description

I've added a new `EncounterPokemonDetail` model along with `encounter_pokemon_details.csv` to add in additional (optional) fields to describe the Pokemon in certain encounters. For now, this includes information like guaranteed perfect IVs, whether or not it is always shiny/non-shiny, and whether or not the Pokemon is an Alpha (helpful as I begin PLA encounters).

In the future, this could maybe help specify natures or movesets from gift encounters if we want to expand to include that, so I think this file can help with those additional details.

When doing this, I encountered errors making the new migration file due to some problems with the changes to the berry models in #1572, so I just marked some fields as nullable in the model.

So far, I've just added details for Pokemon that are always/never shiny to the new `encounter_pokemon_details.csv` file. Checking local builds show that it seems to work and lists the `pokemon_details` field as a list of `dict` objects, just after `encounter_conditions`. If there are any suggestions on the structure otherwise, let me know.

In adding these details, I found some duplicate encounters in SM/USUM and an unexpected Xerneas/Yveltal encounter in the wrong game, so I removed those from the `encounters.csv` file.

No AI was used.

## Contributor check list


data/v2/csv/encounter_pokemon_details.csv
```diff
@@ -0,0 +1,146 @@
+encounter_id,min_perfect_ivs,always_shiny,never_shiny,is_alpha
```
Naramsim, line 1: Instead of always shiny/never shiny we could use a percentage number? like `shiny` and then a value ranging from 0 to 100 or from 0 to 1?

Naramsim, line 1: It's the first time we explicity add IVs in our API. I heard about them but not sure how they are calculated, I think there's a difficult formula. 

questions, can these IVs be calculated with our API? Second, how did you calculate the values you added in the CSV?

jemarq04: IVs aren't calculated, they're used in a calculation for a Pokemon's current stats. Each IV can be 0-31 (with "perfect" being 31). At some point, certain encounters (mostly legendaries) started having guaranteed perfect IVs. I don't know exactly which ones do or how many, but I figured this would be the location to add them.

All I did was go through https://www.serebii.net/games/shiny.shtml to see what Pokemon were guaranteed to be shiny and which were shiny-locked, and then checked each location area in the API to see if we had it already. If we did, I added the enconuter ID and marked it as either always or never shiny.

jemarq04: I think that may be confusing, since this field would need to convey three things: (1) it is always shiny, (2) it's never shiny, or (3) it has the default shiny odds based on the generation/shiny charm/shiny powers and so on. I worry that having something like `shiny` or `shiny_rate` could be confusing if it was `None`, `0`, or `1`. That's why I figured two boolean fields would make it clearer - if both are false, then it's just the usual shiny calculation.
What do you think?

Naramsim: I see! It makes sense. And for calculating the normal shiny rate do people use a formula? Or is it a fixed value? We could add that value here?

Naramsim (approved):

data/v2/csv/encounter_pokemon_details.csv
jemarq04: The value is different depending on the game, and each game has methods of increasing shiny odds like with a shiny charm, sandwiches in SV, donuts in PLZA, and so on. I think documenting each encounter’s shiny rate may be a bit overkill lol

Naramsim: Ok :) just to know :)

pokeapi-machine-user:

pokeapi-machine-user:

PR #1637, opened 2026-08-09
Feat/type checker

FallenDeity:
# Change description

I have written a detailed write up of the changes and decisions made for this pr in the gist u can find here and supporting details, please do give it a read summarizes all the changes nicely
https://gist.github.com/FallenDeity/aa36487a214ea4c40809949a6ccffe7b

here is a tldr for a quick overview

## TLDR

This PR introduces critical performance optimizations, type safety, and linting standards to the codebase:
* performance boosts across the board with serializer and django orm query optimizations with the ditto process completing in 8mins compared to 26~ mins in circleci currently more benchmarks in the gist and optimizations done https://app.circleci.com/pipelines/github/PokeAPI/pokeapi/2617/workflows/729700bf-568b-4cf7-8566-0d4399c07ed9/jobs/7002

* verified that new payloads by server are 100% same as before via a local script

image

* ruff lint all files and type out the django api surface with no errors reported, build and test py files left for future todo to limit scope of typing

Given there is some logical changes i would encourage checking it out locally perhaps running it too for better reviewing, I have tested the pr extensively and there shouldnt be any issues but feel free to request changes, the test and build process havent been modified only logical changes are to serializer.py and tests are running fine so this dosent introduce any regressions

Used some AI for analysis of serializers, some more exploration about django serializers and scripting/visualization to facilitate testing and diff across dumps

## Contributor check list


.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,93 @@
...
+  changed:
+    name: Detect changed files
+    runs-on: ubuntu-latest
+    outputs:
+      run: ${{ steps.filter.outputs.run }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - name: Check changed files
+        uses: dorny/paths-filter@v3
```
Naramsim: Can we get rid of the third party dependency and instead use GHA `paths`? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,93 @@
+name: Code Quality
+
+on:
+  pull_request:
+  push:
+    branches-ignore:
+      - master
```
Naramsim, line 13: This is just a temporary skip, when the code will be all cleaned up we need to remove these two lines?

.github/workflows/database.yml
```diff
@@ -2,6 +2,10 @@ name: Database
 
 on:
   pull_request:
+  push:
+    branches-ignore:
```
Naramsim, line 6: Why ignoring here?

config/__init__.py
Naramsim, line 1: Is this still needed? Can we remove it?

config/asgi.py
```diff
@@ -0,0 +1,16 @@
+"""
+ASGI config for PokeAPI.
```
Naramsim, line 2: This enables ASGI by default? This is why the codebase is faster?

config/docker.py
```diff
@@ -24,3 +26,7 @@
 }
 
 DEBUG = True
+
+for template in TEMPLATES:
```
Naramsim, line 30: Do we need this? Where is TEMPLATE defined?

config/settings.py
```diff
@@ -1,14 +1,59 @@
 # Production settings
 import os
 from pathlib import Path
+from typing import TypedDict
+
+
+class DatabaseSettings(TypedDict, total=False):
+    ENGINE: str
```
Naramsim, line 10: Here do we need to define the struct by ourself or is it possible to import it from some library? Ideally the postrgres and redis one.

config/settings.py
```diff
@@ -91,25 +136,25 @@
     "rest_framework",
     "cachalot",
     "drf_spectacular",
-) + CUSTOM_APPS
+    *CUSTOM_APPS,
```
Naramsim: Can we remove the CUSTOM_APPS and directly add `pokemon_v2` to the list?

config/settings.py
```diff
@@ -155,7 +200,13 @@
     "TAGS": [
         {
             "name": "pokemon",
-            "description": "Pokémon are the creatures that inhabit the world of the Pokémon games. They can be caught using Pokéballs and trained by battling with other Pokémon. Each Pokémon belongs to a specific species but may take on a variant which makes it differ from other Pokémon of the same species, such as base stats, available abilities and typings. See [Bulbapedia](http://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_(species)) for greater detail.",
+            "description": (
+                "Pokémon are the creatures that inhabit the world of the Pokémon games. They can be caught using"
```
Naramsim: Here can we not split the long line? It's because of the linter rules? I find it hard to read by splitting the line over multiple lines.

config/settings.py
```diff
@@ -171,47 +222,67 @@
         },
         {
             "name": "berries",
-            "description": "Berries can be soft or hard. Check out [Bulbapedia](http://bulbapedia.bulbagarden.net/wiki/Category:Berries_by_firmness) for greater detail.",
+            "description": (
+                "Berries can be soft or hard. Check out"
```
Naramsim: Especially the space at the start of the line. I'd leave a single line in these files honestly.

data/v2/build.py
```diff
@@ -787,11 +788,9 @@ def csv_record_to_objects(info):
     build_generic((MoveFlavorText,), "move_flavor_text.csv", csv_record_to_objects)
 
     def csv_record_to_objects(info):
-        _move_effect = None
-        try:
-            _move_effect = MoveEffect.objects.get(pk=int(info[6])) if info[6] != "" else None
-        except:
-            pass
+        move_effect = None
+        with contextlib.suppress(BaseException):
```
Naramsim: Online I see this suppresses also `KeyboardInterrupt` and `SystemExit`. It's a dangerous pattern, it seems.

pokemon_v2/migrations/0032_alter_pokemonevolution_time_of_day.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 5.2.10 on 2026-08-07 19:23
+
+from django.db import migrations, models
```
Naramsim, line 1: Can we have these migrations merged as a single one?

Naramsim (commented):
Hi Fallen! Thanks for the big and important PR! I reviewed some files and asked some questions. 

Note: I didn't review nor take a glimpse at openapi.yml, tests.py, serializers.py, models.py, api.py yet.

.github/workflows/code-quality.yml
FallenDeity: yep, edited to use paths

.github/workflows/code-quality.yml
FallenDeity: well my main thought process for both code-quality and database workflows ignoring master and staging because both are protected branches and we only ever merge from prs so running them in push too seems duplicate running of the same steps, i can enable it if you feel we should run them in both cases

.github/workflows/database.yml
FallenDeity: mentioned above

config/asgi.py
FallenDeity: no its just a config file option to allow people to also run it via asgi aka `uvicorn` over just `gunicorn` which is linux only. its not asgi that makes the codebase faster but the serializers query optimization and using a proper threadpooled client, the benchmarks in the gist compare wsgi and asgi performance with asgi slightly edging out on some endpoints and wsgi on others so rather its not that one posts clearer gains over others but allowing users to run locally using either interface or on prod if they are looking to run pokeapi locally

the difference between asgi vs wsgi is that gunicorn here uses n workers/cores with fixed n threads in our case so if a workload is cpu bound or lot of computation it might perform better than asgi while asgi is just n workers with no limit on threads it context switches in its own event loop so more network bound workloads it would perform better, just giving users more options here our main docker image still uses gunicorn however

config/docker.py
FallenDeity: `TEMPLATE` is defined in settings.py which is our base settings and we have custom config files which override based on runner/env 

https://github.com/FallenDeity/pokeapi/blob/c5cc15b6fc9c29391e111a6f15a94c17f692406e/config/settings.py#L157

the previous `TEMPLATE_DEBUG` was now moved from django settings and now we need to define it in `options` so this sets that for cases where `TEMPLATE_DEBUG` was being set previously

https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/

config/__init__.py
FallenDeity: it marks the config folder as a python package so its necessary and good python convention to follow, also we use it often as `config.settings` or namespace to gather our config modules so this serves as an indicator and some ide's, tools might rely on it too

config/settings.py
FallenDeity: unfortunately django types/stubs packages dont define any backend specific stubs due to wide variance between backend solutions like sqlite, postgres, redis etc so we would need to define specific to our use case

config/settings.py
FallenDeity: yep

data/v2/build.py
FallenDeity: looks like i missed that lint change when committing i think we can improve the look up here using an if check instead, ill fix it

config/settings.py
FallenDeity: yeah its running over 120 length limit, ill revert it to the old format and disable the long line length rule

pokemon_v2/migrations/0032_alter_pokemonevolution_time_of_day.py
FallenDeity: done

config/settings.py
FallenDeity: fixed

data/v2/build.py
FallenDeity: on a sidenote the logic of the build looks wrong?

https://github.com/PokeAPI/pokeapi/blob/17dd3092872cabcb7c008051771d2a2fd8c8c260/data/v2/build.py#L789-L807

index 6 is `priority` not `effect_id` so it looks like wrong values are being set 

for example recoil effect for double edge is in the csv file https://pokeapi.co/api/v2/move/38/ but it dosent reflect in the api and follow me has a chance to poison https://pokeapi.co/api/v2/move/follow-me/

will open a separate issue for this

.github/workflows/code-quality.yml
Naramsim: The code quality it's perfect only on PR. We can skip alltogheter the branches. For the database as well. So 

```
on:
  pull_request:
```

Should be sufficient

FallenDeity: I would still like it to run on push though that way I can see if code quality and database pass in my feature branch rather than raising a pr before i see the workflows running, having push too allows me to verify end to end before raising a pr

data/v2/build.py
Naramsim: True! Some values should be shifted! I'll let you open the PR since you've found the bug! Thanks!

.github/workflows/code-quality.yml
Naramsim: Oh Yeah! Let's keep it this way then :)

pokemon_v2/migrations/0032_evolution_condition_and_form_updates.py
Naramsim, line 1: Sorry maybe I explained myself poorly. Can we leave the existing migrations as they are (except for the sorting of imports) and have a single new `0034` migration?

pokemon_v2/api.py
```diff
@@ -90,12 +173,11 @@ def get_object(self):
...
 @extend_schema_view(list=extend_schema(parameters=[q_query_string_parameter]))
-class PokeapiCommonViewset(ListOrDetailSerialRelation, NameOrIdRetrieval, viewsets.ReadOnlyModelViewSet):
+class PokeapiCommonViewset(ListOrDetailSerialRelation, NameOrIdRetrieval, viewsets.ReadOnlyModelViewSet[Any]):
+    @override
     @extend_schema(parameters=[retrieve_path_parameter])
-    def retrieve(self, request, pk=None):
-        return super().retrieve(request, pk)
-
-    pass
+    def retrieve(self, request: Request, *args: Any, pk: str | int | None = None, **kwargs: Any) -> Response:
```
Naramsim, line 179: Hi, just a curiosity: is the order important here? 

```
self, request: Request, *args: Any, pk: str,
self, request: Request, pk: str, *args: Any, 
```
are the same? Because we have named parameters?

pokemon_v2/api.py
```diff
@@ -114,7 +196,7 @@ def retrieve(self, request, pk=None):
     )
 )
 class AbilityResource(PokeapiCommonViewset):
-    queryset = Ability.objects.all()
+    queryset = Ability.objects.select_related("generation")
```
Naramsim, line 199: Why do we filter for `generation`?

Naramsim: Oh, is this for the N+1 problem?

pokemon_v2/migrations/0032_evolution_condition_and_form_updates.py
FallenDeity: oh ok my bad will fix it tonight

pokemon_v2/api.py
```diff
@@ -887,204 +969,105 @@ class VersionGroupResource(PokeapiCommonViewset):
...
-    },
+    responses={"200": PokeapiMetaResponseSerializer},
 )
 class PokeapiMetaView(APIView):
-    def get(self, request):
+    def get(self, _request: Request) -> Response:
         try:
             git_hash = subprocess.check_output(["git", "rev-parse", "HEAD"], stderr=subprocess.DEVNULL).decode().strip()
-        except Exception:
+        except (subprocess.CalledProcessError, OSError, ValueError):
```
Naramsim, line 1070: Do we need to be so precise here? catching `Exception` is bad?

pokemon_v2/api.py
FallenDeity: not filtering just prefetching/selecting you will notice a similar patter in serializer methods too where we were fetching other models in a for loop over a select_related for example a move might have bunch of version groups fetching the version group in the for loop via its model/serializer makes independent django queries for all related groups instead it can be fetched in one query just by using select_related or pre fetching

Naramsim (commented):
Great, I left some comments/questions. The remaining file that I need to review is serializers.py

pokemon_v2/api.py
FallenDeity: not the same, order is important that keyword arguments should follow positional parameters. 

in case 1 pk is explicitly a keyword argument 
in 2nd its positional u can pass either as position or as kwarg

pokemon_v2/api.py
FallenDeity: well a general rule of thumb is we never almost want a base exception anywhere because say for example subprocess someone entered an extra `s` as typo it would catch that too, basically just ends up masking errors which otherwise couldve been avoided by being specifc. another benefit is by having a specific error its so much clearer to the reader what could possibly go wrong and what cases are accounted for

also ruff lint throws an error when we have a blind exception https://docs.astral.sh/ruff/rules/blind-except/

pokemon_v2/migrations/0032_evolution_condition_and_form_updates.py
FallenDeity: fixed

Naramsim (approved):
Well, what to say. This is incredible. I downloaded the changes on my PC and everything is way faster than before. I took a look at the serializer file and got the gist of it but couldn't review it all. 

For me we can merge this in. I don't know if @phalt or @jemarq04 want to take a look as well.

jemarq04:
I did a couple local tests and skimmed through the changes - this is a lot of great work! Thanks @FallenDeity! It looks like Naramsim was very thorough as well, so I'll trust that check.

Naramsim:
Hey @FallenDeity i merged the PR! I retained the commits as well! 

Thanks a lot for all the speed improvements and code cleaning!

pokeapi-machine-user:

FallenDeity:
@Naramsim looks like deploy failed, an update is needed here 

https://github.com/PokeAPI/api-data/blob/master/updater/cmd.bash#L27-L28

to docker_compose

edit: made a pr fixing this https://github.com/PokeAPI/api-data/pull/422

Naramsim:
Good catch!

pokeapi-machine-user:

pokeapi-machine-user:

FallenDeity:
ggs nice lot of time reduction
```
super-contest-effect (3x): 100%|██████████| 22/22 [00:00<00:00, 91.80it/s]
type (3x): 100%|██████████| 21/21 [00:00<00:00, 28.89it/s] ?it/s]
version (3x): 100%|██████████| 53/53 [00:00<00:00, 70.24it/s]
version-group (3x): 100%|██████████| 32/32 [00:00<00:00, 76.22it/s]
clone: 100%|██████████| 50/50 [09:10<00:00, 11.01s/it]t/s]
+ ditto clone --src-url http://localhost/ --dest-dir ./data --select pokemon/129
```
should be possible to bring this down to 2-3mins with ditto running a thread pool, instead of procs

Naramsim:
That's a great time saving!

PR #1642, opened 2026-08-20
fix: make API list ordering deterministic

MiquelRForgeFlow:
Most ORM queries had no explicit ORDER BY, so the database was free to return rows in any order. The same dataset could then be serialized differently on every build, turning each api-data regeneration into a huge diff of pure reordering noise.

Give PokeApiModel a default manager that orders by primary key. Managers are inherited from abstract bases and reverse relation managers derive from the model's default manager, so this covers plain queries, reverse relations and prefetches at once, including the declarative reverse fields (Generation, GrowthRate, PokemonColor, PokemonShape, PokemonHabitat) that no serializer call site could order.

Add tie-breakers to the partial orderings that left equal keys undefined, ending every order_by() with the primary key: pokemon moves and held items, location area and pokemon encounters, past abilities, stats and types, form and pokemon types, berry flavors, pokedex entries and encounter method rates. The few calls left untouched already order by order_by("id"), which is the primary key.

Document in the manager that a distinct(*fields) call has to spell out a matching order_by(), since PostgreSQL requires the leading ORDER BY expressions to match the DISTINCT ON ones. A plain distinct() is fine, because the primary key is already part of the selected row.

Add two regression tests that enforce the ordering, one for a declarative reverse relation and one for rows tying on the whole ORDER BY. Both insert rows with primary keys out of order, so both fail without this fix.

No migration needed; call sites needing another order still override it.

Of course, the PokeApiManager was AI idea. Also used to check nothing is left out. Used to create tests too.

## Contributor check list

### DIFF
--- a/pokemon_v2/models.py
+++ b/pokemon_v2/models.py
@@ -1,10 +1,12 @@
 # pyright: reportIncompatibleVariableOverride=false
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, ClassVar, TypeVar
 
 from django.db import models
-from typing_extensions import override
+from typing_extensions import Self, override
+
+_ModelT = TypeVar("_ModelT", bound=models.Model)
 
 __all__: tuple[str, ...] = (
     "Ability",
@@ -165,6 +167,7 @@ __all__: tuple[str, ...] = (
     "PalPark",
     "PalParkArea",
     "PalParkAreaName",
+    "PokeApiManager",
     "PokeApiModel",
     "PokeathlonStat",
     "PokeathlonStatName",
@@ -230,7 +233,24 @@ __all__: tuple[str, ...] = (
 ############################
 
 
+class PokeApiManager(models.Manager[_ModelT]):
+    """Default manager ordering every queryset by primary key, so that the API
+    serializes lists deterministically. Call sites needing another order
+    override it with ``order_by()``.
+
+    A ``distinct(*fields)`` call has to spell out a matching ``order_by()``,
+    since PostgreSQL requires the leading ``ORDER BY`` expressions to match the
+    ``DISTINCT ON`` ones. A plain ``distinct()`` is unaffected.
+    """
+
+    @override
+    def get_queryset(self) -> models.QuerySet[_ModelT]:
+        return super().get_queryset().order_by("pk")
+
+
 class PokeApiModel(models.Model):
+    objects: ClassVar[PokeApiManager[Self]] = PokeApiManager()
+
     class Meta:
         abstract = True
 
--- a/pokemon_v2/tests.py
+++ b/pokemon_v2/tests.py
@@ -1309,6 +1309,7 @@ class APIData:
         is_legendary=False,
         is_mythical=False,
         order=1,
+        pk=None,
     ):
         generation = generation or cls.setup_generation_data(name="gen for " + name)
 
@@ -1321,6 +1322,7 @@ class APIData:
         pokemon_habitat = pokemon_habitat or cls.setup_pokemon_habitat_data(name="pkm hbtt for " + name)
 
         pokemon_species = PokemonSpecies.objects.create(
+            pk=pk,
             name=name,
             generation=generation,
             evolves_from_species=evolves_from_species,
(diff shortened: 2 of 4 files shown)


Naramsim:
Hi @FallenDeity Can you chime in and review this PR? Maybe I'm wrong but your current re-implementation is already preserving a custom-order of the fields, am I wrong? On my PC when building the current `master` branch I always get the same results/order. 

If this PR enforces the order or a specific sorting strategy we can merge it in! 

@MiquelRForgeFlow Thanks for the PR!

Naramsim (approved):
I trust the tests on this PR. I checked it out locally and everything seems to run

phalt:
I remember trying to do this years ago. One of those ones where you see a solution and realise "damn, it was that easy?" Will make reviewing the pokeapi-data diffs in the future far easier 🙏

FallenDeity:
> Hi @FallenDeity Can you chime in and review this PR? Maybe I'm wrong but your current re-implementation is already preserving a custom-order of the fields, am I wrong? On my PC when building the current `master` branch I always get the same results/order. 
> 
> If this PR enforces the order or a specific sorting strategy we can merge it in! 
> 
> @MiquelRForgeFlow Thanks for the PR!

i did not do anything for order in my prior pr it was mainly just serializer query optimization, I have a few thoughts on why the same order is maintained few of the serializers already have order_by where now pk has been added as a fallback and also we freshly load the data in sqlite/postgres before serving which might tend to return the rows in same physical order its defined in the csv since there is no db reorg happening 

this change is limited to the nested json lists and makes the behaviour consistent 

like this for example
```
abilities = AbilitySummarySerializer(many=True, read_only=True, source="ability")
```
now it will call the query set on the manager with the ditto change i am working on it should make diff viewing a lot easier yeah

awesome work :)

few reviews
- we should add a test case for this which enforces the order
- wherever order_by is present pk should be as fallback its there in most places, few places no pk any reason for that
```diff
- BerryFlavorMap.objects.filter(berry_flavor=obj, potency__gt=0).select_related("berry").order_by("potency")
+ BerryFlavorMap.objects.filter(berry_flavor=obj, potency__gt=0)
            .select_related("berry")
            .order_by("potency", "berry_id")
        )
```

also add a comment in the manager that if distinct() is used the fields for both order_by() and distinct() must be same https://stackoverflow.com/questions/9795660/postgresql-distinct-on-with-different-order-by

MiquelRForgeFlow:
Thanks for the reviews! All three points addressed in the amended commit:

- `pk` added to the seven `order_by()` calls missing it. The three left alone already order by the primary key via `order_by("id")`.
- `distinct()` caveat documented in the manager docstring.
- Two regression tests, both verified to fail without the fix (`[2, 1, 3] != [1, 2, 3]`).

pokemon_v2/tests.py
```diff
@@ -4591,6 +4620,40 @@ def test_pokemon_moves_with_non_contiguous_version_group_and_method_ids(self):
             "{}{}/move-learn-method/{}/".format(TEST_HOST, API_V2, pokemon_move.move_learn_method.pk),
         )
 
+    def test_pokemon_moves_version_group_details_are_deterministically_ordered(self):
+        # Regression test: get_pokemon_moves only ordered by move and
```
FallenDeity: going in the future this comment would be come redundant or confusing since this issue is no longer the case once its merged, you can just write a one liner saying worst case `order_by` falls back to `pk`

pokemon_v2/tests.py
```diff
@@ -2686,6 +2689,32 @@ def test_growth_rate_api(self):
             "{}{}/pokemon-species/{}/".format(TEST_HOST, API_V2, pokemon_species.pk),
         )
 
+    def test_reverse_relation_lists_are_ordered_by_pk(self):
```
FallenDeity, line 2692: same here too, its kind of like a narration of the issue or context

pokemon_v2/models.py
```diff
@@ -230,7 +233,32 @@
 ############################
 
 
+class PokeApiManager(models.Manager[_ModelT]):
+    """Default manager that guarantees a stable, reproducible row order.
+
+    Without an explicit ``ORDER BY`` the database is free to return rows in any
```
FallenDeity: just shorten this saying the django queryset manager class here defaults to sort or ordering by pk to be deterministic

FallenDeity:
reviewed lgtm left a few suggestions regarding the comments give that a look when ur free should be good to merge

MiquelRForgeFlow:
Good point, all three comments were narrating the bug rather than the code. Shortened them.

Naramsim:
Thanks for the code @MiquelRForgeFlow and thanks for the review @FallenDeity ! 

PS: @FallenDeity I invited you to the PokeApi's GitHub organization! Thanks so much for all the efforts you've put in over the last years!

pokeapi-machine-user:

pokeapi-machine-user:

PR #1649, opened 2026-08-25
Update GitHub issue templates

jemarq04:
# Change description

I've updated the bug report issue template for GitHub so that it now uses the form submission. I've attached an image here to show what it would look like - it's a bit more structured now and includes a comment on opening a PR or asking others for help. (The other option not shown in the bottom dropdown is "I plan on making a PR for this issue." The issue that is created will look the same as it had been previously with the markdown template.
bug-report-form

I've also updated the contact links to remove the old slack/stackoverflow links to the new Discord server. (That being said, we may need to update the invite since we want to point the invite to the `#welcome` channel instead of `#general`).

I'm open to suggestions on these updates! If you think anything else would improve readability I can add that in.

No AI was used.

## Contributor check list


jemarq04:
Hmm. Maybe we want to add path filters to some of the CI - no need to run all of our tests a PR that only updates `.github/`.

phalt (approved):
Woah, so cool we can do this now!

jemarq04:
Yeah, it seemed like a neat feature so I thought we could try it out! For now I left the feature requests one alone. There is also an auto-labeling feature in the template, but I wasn't sure how we felt about that since we don't have those in use right now

Naramsim (approved):

Naramsim:
Yeah! Adding more structure will help contributors, I also like a lot the suggestion about opening a PR

pokeapi-machine-user:

PR #1652, opened 2026-08-27
feat: add french translations for locations and encounter methods

Lafoxnir:
# Change description

Added new french translations for encouter methods and location names. Almost everything is done up to Sword and Shield, which are a little harder to translate because of max dens (see below).

I also fixed some typos in the locations and added missing english encouter methods (which means they all have an entry in encounter_method_prose.csv)

About Max Dens and their translations :
- Each Max Den is considered a separate location in PokeAPI (which is fair ince Pokémon pools are different) and use the Bulbapedia order for the letters
- However, on Poképedia (the french equivalent of Bulbapedia), the Max Dens are not in the same order
- Since Poképedia is accepted as an official source for french data, I wanted to use this order for the translations, so a Max Den A in english might not be A in French
- For example see the Rolling Fields [in English](https://bulbapedia.bulbagarden.net/wiki/Rolling_Fields/Dens) and [in French](https://www.pokepedia.fr/Plaine_Verdoyante/Antres), where Max Den A is D in French
- For the base game maps, I could use the minimap to help me a lot
- For Isle of Armor, however, the minimap is gone in English (still here in French) so I used the image and description of each Max Den to compare, except when arriving at some point where even the Max Den image was gone in English (so comparison is too hard)
- For Crown Tundra, there is no map, no Max Den image, and no location description, so I cannot compare anything

So the two solutions I see are :
- leaving this as I did to keep the order in French (meaning some locations might never get translated)
- modifying everything to keep the same order in French than in English (which means that French users will not have their official order (again, according to Poképedia))

Let me know if you have some suggestions for this

No AI was used

## Contributor check list

### DIFF
--- a/data/v2/csv/encounter_method_prose.csv
+++ b/data/v2/csv/encounter_method_prose.csv
@@ -78,8 +78,18 @@ encounter_method_id,local_language_id,name
 30,5,En utilisant le Devon Scope
 31,9,Fishing in a Feebas Tile
 31,5,En pêchant sur une tuile à Barpau
+32,9,Island Scan
+32,5,Scan des Îles
+33,9,SOS Battle
+33,5,Combat SOS
 34,9,Fishing at bubbling spots in the water
+34,5,En pêchant sur un point à bulles
 35,9,Interacting with a Berry Tree
+35,5,En interagissant avec un arbre à baies
+36,9,Trade with an NPC
+36,5,Échange avec un PNJ
+37,9,SOS Battle from bubbling spots
+37,5,Combat SOS dans les points à bulles
 38,9,Walking in the overworld
 38,5,En marchant dans le monde extérieur
 39,9,Surfing in the overworld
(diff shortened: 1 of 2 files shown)


Naramsim (approved):

Naramsim:
Thanks!

pokeapi-machine-user:

Naramsim:
@Lafoxnir I might have merged a bit to early. But I think it's ok. The letters for the Dens are language specific I think. So if someone is playing in English they might see a letter A and for the same location a french player might see D, am I right? If this is what the game presents to the user then we are good to go with this PR. Else, if those letters are came up with by the creators of Bulbapedia/Pokepedia then I'm unsure about what to do. I'd me more inclined to have the names consistent, but then the French players relying on pokepedia will be lost , right?

pokeapi-machine-user:

Lafoxnir:
@Naramsim Yes that's right, the letters are language specific and there does not seem to be an official naming convention, nothing in name

For the French letters, it seems to be (per zone) from the north-western corner to the south-earthern corner of each map while scanning the map horizontally. We also have numbers (https://www.pokepedia.fr/Antre_1) that seem to correspond to Serebii numbers (https://www.serebii.net/swordshield/maxraidbattles/den1.shtml) but I don't know where these numbers come from

I can't seem to understand the naming convention from Bulbapedia but maybe it is following the Serebii numbers in ascending order for each zone?

Naramsim:
We could then create our numbering convention so to make things even more difficult 🤩. Jokes apart I think there's no solution here since we don't even host a map where to see those letters. 

I'd use a single numbering system and I'm leaning towards the easier to understand, maybe even adding a description in the docs. I wouldn't like to link the map hosted at Bulbapedia so, as the French wiki is doing, from the top left to the bottom right seems understandable enough? Or maybe Bulbapedia and all the other English wikis use the same different ordering system and users will be left puzzled? 

@jemarq04 what do you think?

